PR #13, opened 2015-12-24
(SIMP-392) Removed simp_enabled fact

trevor-vaughan:
- simp_enabled is not needed on the client side
- Restored the old acceptance tests

SIMP-392 #comment Corrected issues found during testing


trevor-vaughan:
Closed via Gerrit f4756fa0616a19fd90188cb4c347d51fe2323f6d

PR #17, opened 2016-02-03
Fix ipv6_enabled

Klaas-:
If you completly disable ipv6 via options ipv6 disable=1 at module load
this fact throws annoying errors during every puppet run
adding -e will not print error and return ""

PS: Sorry for not creating an issue in your jira


op-ct:
Good catch, Klaas.  Thanks for the fix!

PR #34, opened 2016-06-27
(SIMP-705) Install tmpwatch package for EL6 systems

jeefberkey:
We have a cron job for tmpwatch but it may not necessarily exist.

This should only apply on EL6 systems.

SIMP-705 #close


trevor-vaughan:
Merged via Gerrit patches

PR #60, opened 2016-11-27
(SIMP-1694) Added 'simp_opt' function

trevor-vaughan:
The simp_opt function is designed for retrieving SIMP 6-style global
catalysts from any viable source.

SIMP-1694 #close

### DIFF
--- a/lib/puppet/parser/functions/simp_opt.rb
+++ b/lib/puppet/parser/functions/simp_opt.rb
@@ -1,0 +1,49 @@
+module Puppet::Parser::Functions
+  newfunction(:simp_opt, :type => :rkey, :arity => 1, :doc => <<-EOS
+    Retrieve a SIMP global catalyst.
+
+    Any variable that is passed into this function will look up a variable
+    ``simp_opt::<variable>``.
+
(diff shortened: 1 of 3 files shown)


lib/puppet/parser/functions/simp_opt.rb
```diff
@@ -0,0 +1,49 @@
+module Puppet::Parser::Functions
```
jeefberkey: Should we still be creating functions this way?

lib/puppet/parser/functions/simp_opt.rb
```diff
@@ -0,0 +1,49 @@
...
+  newfunction(:simp_opt, :type => :rkey, :arity => 1, :doc => <<-EOS
```
op-ct: In Puppet 4, I think we're supposed to use [Puppet::Functions.create_function](https://github.com/puppetlabs/puppet-specifications/blob/master/language/func-api.md#creating-a-function)

trevor-vaughan:
We're not going to be using this so I'm killing the PR.

PR #79, opened 2017-01-27
(SIMP-2547) Add fips_ciphers fact

jeefberkey:
The fips_ciphers fact returns the available OpenSSL FIPS ciphers on the
system. This will be used as the default ciphers set for OpenLDAP,
because it doesn't like the FIPS shortcut.

SIMP-2546 #comment added the fact
SIMP-2547 #close

### DIFF
--- a/lib/facter/fips_ciphers.rb
+++ b/lib/facter/fips_ciphers.rb
@@ -4,6 +4,6 @@ Facter.add('fips_ciphers') do
   confine :kernel => 'Linux'
 
   setcode do
-    Facter::Core::Execution.exec(`openssl ciphers FIPS:!LOW`).split(':')
+    Facter::Core::Execution.exec('openssl ciphers FIPS:!LOW').split(':')
   end
 end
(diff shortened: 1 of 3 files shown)


lib/facter/fips_ciphers.rb
```diff
@@ -0,0 +1,9 @@
+# List available FIPS-compatible OpenSSL ciphers on the system
+# Returns: Array[String]
+Facter.add('fips_ciphers') do
+  confine :kernel => 'Linux'
+
+  setcode do
+    Facter::Core::Execution.exec(`openssl ciphers FIPS:!LOW`).split(':')
```
jeannegreulich: This does not work on CentOS 6.  I had to change it to 'openssl ciphers FIPS:LOW'  changing the ` to ' and getting rid of the !.

jeefberkey: Whoops I forgot to change that. It doesn't work on 7 either.

PR #82, opened 2017-03-06
WIP: (SIMP-2785) Allow passgen to use libkv

heliocentric:
* Add v4 scoped function of 'simplib::passgen' that
  implements passgen using libkv as the backend
* Add an option to the top scoped 'passgen', 'libkv',
  which defaults to the value of simp_options::libkv.
  When true, it will call simplib::passgen

SIMP-2785 #close


lib/puppet/functions/simplib/passgen.rb
```diff
@@ -0,0 +1,127 @@
+# vim: set expandtab ts=2 sw=2:
```
jeefberkey: Documentation here

metadata.json
```diff
@@ -15,6 +15,10 @@
     "alias"
   ],
   "dependencies": [
+     {
```
jeefberkey: Should this requirement be added now?

heliocentric: Yes, it has to be for the functions to be loaded; dependency in metadata json is used by DependencyLoader when loader.load(:function, "libkv::something") is called by call_function

jeefberkey: Oh yeah, I forgot about that. I was just thinking if users didn't have the catalyst set. Adding a dep to simplib means it will have to be on literally every machine using any simp module.

heliocentric: yea, and it doesn't do anything but lay down the functions. It will be required for 7.x

lib/puppet/functions/simplib/passgen.rb
```diff
@@ -0,0 +1,127 @@
...
+        passwd = passwd[0..(length-1)]
+      end
+    rescue Timeout::Error
+      raise Puppet::ParseError, "passgen timed out for #{@id}!"
+    end
+
+    return passwd
+  end
+
+  def passgen(id, hash = nil)
```
lamawithonel: This whole thing needs more blank lines and comments.

lib/puppet/functions/simplib/passgen.rb
```diff
@@ -0,0 +1,127 @@
...
+    end
+
+    passwd = ''
+    begin
+      Timeout::timeout(30) do
+        default_charlist = ("a".."z").to_a + ("A".."Z").to_a + ("0".."9").to_a
+        specific_charlist = nil
+        case complexity
+        when 1
+          specific_charlist = ['@','%','-','_','+','=','~']
```
lamawithonel: Please add spaces between these elements. It's quite difficult to read like this.

heliocentric:
This new behavior should be documented and advertised in the README. I think passgen is an underappreciated part of SIMP and this is a great opportunity to make it a little more obvious.
- Yep, I'm going to update a lot of docs (as well as puppet-strings and inline comment) this before removing WIP.

I see that you modified the old function to use the new function if the libkv catalyst is set. That should make our existing code work just if that setting is there.
- Yep; for the 6.x branch passgen will default to the same behaviour. 7.x will switch it to on by default.

What if you have existing passwords laying around in the old simp_autofiles dir? 
- So, the goal is to create a libkv backend called simp6-legacy, which will emulate the simp_autofiles behaviour within the libkv api. ie; if you set libkv = true, and libkv::url is simp6-legacy://; it should act 100% the same as if libkv == false. Then, the plan is to have a migrate puppet class class in libkv, that just does this psuedo-code:
libkv::list("/").each do |name|
  value = libkv::get(name)
  libkv::put(name, value, { url => newurl})
end
and syncs the current libkv location to the newurl location. Then you can switch libkv::url to newurl and automagically convert.

How do you bootstrap SIMP without a libkv compatible service being up?
- SIMP config/bootstrap will eventually have something for libkv so that either simp6-legacy or a generic file:// backend is set up. or just always consul. That decision hasn't really been made and will take some coordination

lib/puppet/functions/simplib/passgen.rb
```diff
@@ -0,0 +1,127 @@
...
+          ]
+        end
+
+        charlists.each do |charlist|
+          (length/charlists.length).ceil.times { |i|
+            passwd += charlist[rand(charlist.size-1)]
+          }
+        end
+
+        passwd = passwd[0..(length-1)]
```
lamawithonel: As long as you're here, this might as well be cleaned up.  http://stackoverflow.com/a/88341/2044979

This should probably use `SecureRandom`, as well.

lib/puppet/functions/simplib/passgen.rb
```diff
@@ -0,0 +1,127 @@
...
+        unless (stored_pass["value"].length == length)
+          retval = attempt_put = call_function("libkv::atomic_put", { "key" => password_path, "value" => pass, "previous" => stored_pass})
+        end
+      end
+      if (retval == true)
+        successful_put = true
+        break
+      end
+    end
+    if (successful_put == false)
```
lamawithonel: This should probably fail for safety. Returning `nil` for a password or returning some value which has not been stored could both be very bad.

nick-markowski:
I'd like to leave this PR open until we decide mine (https://github.com/simp/pupmod-simp-simplib/pull/123) is not garbage.

lnemsick-simp:
libkv prototype 2 changes have already been made to simplib::passgen and the points in this discussion have been entered as libkv tickets.

PR #90, opened 2017-04-07
(SIMP-2994) Use uppercase for LDAP DN strings in simplib::ldap::domain_to_dn

heliocentric:
* Use uppercase 'DC=' instead of lowercase 'dc=' so systems like active
  directory can handle the output.

SIMP-2994 #close


trevor-vaughan:
@heliocentric Rebase please, otherwise looks OK.

Did you check that this doesn't break the other tests (it shouldn't)

PR #92, opened 2017-04-28
(SIMP-3115) Added wrapper stages to simplib

trevor-vaughan:
There have been several situations where various SIMP components really
needed to be wrapped in a Puppet stage to protect the rest of the
system from a particularly aggressive action (usually one that requires
a reboot).

This change adds a couple of SIMP stages that wrap the stdlib stages to
provide a buffer zone for SIMP-specific activities. This is not a
perfect solution but it will work better than just trying to tag onto
the existing stages that don't actually map to our use cases.

SIMP-3115 #comment Added buffer stages for module locking

### DIFF
--- a/lib/facter/simplib_sysctl.rb
+++ b/lib/facter/simplib_sysctl.rb
@@ -1,0 +1,40 @@
+# _Description_
+#
+# Returns a hash of sysctl values that are particularly relevant to SIMP
+#
+# We don't grab the entire output due to the sheer size of it
+#
+Facter.add("simplib_sysctl") do
+  setcode do
+    relevant_entries = [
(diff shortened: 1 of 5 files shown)


lib/facter/simplib_sysctl.rb
```diff
@@ -0,0 +1,46 @@
+# _Description_
```
jeefberkey: @trevor-vaughan you said you had a better way to do this. It should go into another PR and ticket.

trevor-vaughan: This *is* the better way.

The *other* way is to dump all of sysctl into a hash, but I don't think we want that madness in every `facter` run.

PR #93, opened 2017-05-05
(SIMP-3138) Added new root_dir_uuid fact

trevor-vaughan:
* Cleaned up the boot_dir_uuid fact
* Reverted the version number for the module since the last release was
  3.2.0

SIMP-3138 #comment Added supporting root_dir_uuid fact for the fips module


trevor-vaughan:
Killing this because @jeefberkey is right :-D

trevor-vaughan:
@jeefberkey Re-opening this. The `partitions` fact doesn't actually tell you the information for the `/` partition, which is what we need.

PR #102, opened 2017-07-17
(SIMP-3437) Add assert_metadata_os function

trevor-vaughan:
Added a function that allows users to easily fail a compile if the
client's OS does not match an OS in the metadata.json of the passed
module.

SIMP-3437 #close


functions/assert_metadata_os.pp
```diff
@@ -0,0 +1,57 @@
+# Fails a compile if the client system is not compatible with the module's
```
jeefberkey: Why do we need to explicitly fail if a particular module doesn't support an OS? We already codify that we only support certain OSs in our metadata.json and in simp-core/the iso. Why should we punish those who try different ones?

trevor-vaughan: @jeefberkey This replaces the logic that used to be all over `params.pp` where you had a catch statement that failed on unsupported Operating Systems. Now, instead of a big block of logic, you can just call assert and have it take care of it for you.

jeefberkey: Fair.

functions/assert_metadata_os.pp
```diff
@@ -0,0 +1,57 @@
...
+#   The name of the module that should be checked
+#
+# @param match
+#   The level at which to match the release number
+#
+#   * none  -> No match on release (default)
+#   * full  -> Full release must match
+#   * major -> Only the major release must match
+#
+function simplib::assert_metadata_os (
```
jeefberkey: Can the function be called something a little less ambiguous? Maybe:
* `validate_supported_os`
* `on_supported_os`
* `check_metadata_for_os`

trevor-vaughan: @jeefberkey `check_metadata_for_os` is the least ambiguous of these but not really better than what's there. `assert` means fail if bad and `metadata_os` isn't very ambiguous.

I started by calling it `validate`, but that turns into something like `validate_os_from_metadata` which I don't really think is any better than what is already there.

lamawithonel: I think it should be `assert_os_support()`, because it is an assert function in the classic sense; it is also generic.

My biggest thing is this: the function name shouldn't know about the implementation, nor should the classes using this function.

Just imagine it like this: you add this function to a hundred modules, then sometime down the line you improve it so it can use some new feature to determine OS compatibility. You could keep using the same function name, but then doesn't make sense and creates cognitive dissonance in users who are trying to figure out why the function is failing their modules. You could also wrap the old function in a new one, but then you have all these old modules using the old function, not gaining the benefits of the new one.  Or, god forbid, you could go change the name in all those hundred plus odd modules-- sound familiar? On the other hand, you could start right now by using a generic name, then not have to worry about any of this later on.

op-ct:
@lamawithonel, it's safer to `fail()` by default if a module is applied on an unsupported OS.  It's also the expected behavior.  The [Puppet style guide](https://docs.puppet.com/puppet/4.8/style_guide.html#defaults-for-case-statements-and-selectors) has this blurb on it:

> When setting the default case, keep in mind that the default case should cause the catalog compilation to fail if the resulting behavior cannot be predicted on the platforms the module was built to be used on.

Until recently, the logic for this was usually [kept in `::params.pp`](https://github.com/garethr/puppet-module-skeleton/blob/master/skeleton/manifests/params.pp.erb#L16-L17).  However, the `::params` hack was retired after data-in-modules.  This function re-implements the chunky `::params` logic into a simple one-liner that is easier to understand and maintain.

I disagree with the proposed (meta-)parameter toggle, because _convention over configuration._

functions/assert_metadata_os.pp
lamawithonel: Per @op-ct's comment, how about `assert_supported_configuration()`?

lamawithonel:
Good point. Convention over configuration would dictate that it not include a function parameter, but that we establish some convention in how it is used. Per that, I still like the documentation idea suggesting class parameter to disable it.

op-ct:
I think the class (meta-)parameter is still configuration over convention, among other things.  

Most popular Forge modules follow the convention of `fail()`ing on unsupported environments, as anyone who risks applying one to an unsupported OS will quickly learn.

It's reasonable to expect such users to be aware of this convention—and to be prepared to work around it on their own copy as they would with any other module (and probably this one's dependencies).  It seems less reasonable to shift that burden onto each module by way of an unconventional configuration (meta-)parameter and accompanying puppet logic.

op-ct:
For the function's convention: I'd propose that is should be placed near the top of the class in each `init.pp` as a single-statement gatekeeper, similar to how `assert_private()` is used in private classes.

lamawithonel:
@op-ct  I well aware of the style guide's recommendation, and I've experienced the pain of a half-finished run, as well. However, I've also experienced cases where module developers were overly strict about which systems or versions of things they supported. In those cases, the modules worked just fine on the "unsupported" systems, but due to the developer's policy-- their inability or unwillingness to spend time running tests for systems they did not use-- they would not accept patches adding a line to their `params.pp` or `metadata.json` file, even if it included tests. That meant I had to maintain a fork of their module with all the overhead that entails.  If there were an easy way to disable the supported configuration check, however, then I could have just set a key Hiera and been done with it-- continued using the module just like any other third-party module.

On your other point, the options hash and too much surrounding logic would be a little too much. I'm glad you pointed that out. Maybe better would be to make the Boolean toggle an optional first argument, then it could be used like this:

```puppet
class foo (
  Boolean $fail_unsupported_config = true,
) {
  assert_supported_config($fail_unsupported_config)

  ...
```

If set to false, it could warn instead of fail, stating that any warranties or support contracts associated with the module may be null and void until the systems is brought back into a supported configuration. Otherwise, it fails hard with the usual output, per current best practices.

trevor-vaughan:
@lamawithonel @op-ct @jeefberkey The latest push should satisfy everyone and make the usage horribly overcomplex 😑

functions/assert_metadata.pp
```diff
@@ -0,0 +1,95 @@
+# Fails a compile if the client system is not compatible with the module's
+# `metadata.json`
+#
+# @param module_name
+#   The name of the module that should be checked
+#
+# @param options
+#   Behavior modifiers for the function
+#   * Can be set using `simplib::assert_metadata::options` in the `lookup`
+#     stack
```
lamawithonel: If I use this in a bunch of modules in my environment, and I want to test just one of them against an unsupported configuration, would setting this option disable it for everything?

trevor-vaughan: @lamawithonel Yes, it would disable it for everything, doing otherwise would be even more overly complex.

trevor-vaughan:
@lamawithonel @op-ct @nick-markowski Can someone re-review this? Old reviews are killed by PRs now.

PR #105, opened 2017-07-31
(SIMP-3434) Added slapd_version fact

nick-markowski:
Added slapd_version fact. Primary use is in the simp_openldap module,
which configures TLS options based on the version of slapd.

SIMP-3434

### DIFF
--- a/lib/facter/slapd_version.rb
+++ b/lib/facter/slapd_version.rb
@@ -1,0 +1,17 @@
+# Set a fact to return the version of slapd that is installed
+#
+Facter.add("slapd_version") do
+  if ['RedHat','CentOS'].include?(Facter.value(:operatingsystem))
+    if Facter.value(:operatingsystemmajrelease) < '7'
+      $slapd_bin = '/usr/sbin/slapd'
+    else
+      $slapd_bin = '/sbin/slapd'
+    end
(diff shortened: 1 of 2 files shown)


lib/facter/slapd_version.rb
```diff
@@ -0,0 +1,17 @@
+# Set a fact to return the version of slapd that is installed
+#
+Facter.add("slapd_version") do
+  if ['RedHat','CentOS'].include?(Facter.value(:operatingsystem))
+    if Facter.value(:operatingsystemmajrelease) < '7'
```
jeefberkey: `Facter::Core::Execution.which('slapd')` will abstract this

lib/facter/slapd_version.rb
```diff
@@ -0,0 +1,17 @@
+# Set a fact to return the version of slapd that is installed
+#
+Facter.add("slapd_version") do
+  if ['RedHat','CentOS'].include?(Facter.value(:operatingsystem))
```
jeefberkey: This isn't needed if you use `confine` properly

nick-markowski: Ill fix the confine

nick-markowski:
@jeefberkey I forgot I was working with this fact in a profile module, I'll move it

PR #106, opened 2017-08-03
(SIMP-3490) Updated the fips_ciphers fact

nick-markowski:
- Returns nil if the openssl binary is not available
- Addresses https://github.com/simp/pupmod-simp-simp_grafana/issues/28

SIMP-3490 #close


nick-markowski:
@jeefberkey Good question. simp_options already has logic and tests in place to handle fips_ciphers = nil. See simp_options/manifests/openssl/params.pp and simp_options/spec/classes/openssl_spec.rb.

PR #109, opened 2017-08-15
(SIMP-3355) simplib: Deprecate & convert Puppet3 functions part 1

lnemsick-simp:
- Add simplib-specific deprecation functions for both Puppet 3 functions
  (simplib_deprecation()) and Puppet 4 functions (simplib::deprecation()).
- Convert a subset of Puppet 3 functions to Puppet 4 and emit a
  simplib deprecation warning when the Puppet 3 versions as called:
  - simplib::inspect() replaces deprecated inspect()
  - simplib::ipaddresses() replaces deprecated ipaddresses()
  - simplib::parse_hosts() replaces deprecated parse_hosts()
  - simplib::passgen() replaces deprecated passgen()
  - simplib::strip_ports() replaces deprecated strip_ports()
  - simplib::to_integer() replaces deprecated to_integer()
  - simplib::to_string() replaces deprecated to_string()
  - simplib::validate_uri_list() replaces deprecated validate_uri_list()
  - simplib::validate_net_list() replaces deprecated validate_net_list()
  - simplib::validate_port() replaces deprecated validate_port()
- Add single-line json output option to simplib::inspect()


lib/puppet/functions/simplib/to_integer.rb
```diff
@@ -0,0 +1,21 @@
+# Converts the argument into an `Integer`.
+# Only works if the passed argument responds to the `to_i()` Ruby method.
+#
+Puppet::Functions.create_function(:'simplib::to_integer') do
```
jeefberkey: For these conversion functions, I think we should encourage use of [assrt_type](https://docs.puppet.com/puppet/latest/function.html#asserttype) or regular type manipulation listed in each data type's docs page ([string](https://docs.puppet.com/puppet/5.0/lang_data_number.html#converting-strings-to-numbers) for example) and the old functions should just be deprecated and abandoned.

lib/puppet/functions/simplib/validate_port.rb
```diff
@@ -0,0 +1,39 @@
+# Validates whether or not the passed argument is a valid port
+# (i.e. between `1` - `65535`).
+Puppet::Functions.create_function(:'simplib::validate_port') do
```
jeefberkey: Has this been replaced by data type Simplib::Port and friends? https://github.com/simp/pupmod-simp-simplib/blob/master/types/port.pp

lib/puppet/parser/functions/to_integer.rb
```diff
@@ -10,6 +10,8 @@ module Puppet::Parser::Functions
     @return [Integer]
     EOS
 
+    function_simplib_deprecation(['to_integer', 'to_integer is deprecated, please use simplib::to_integer'])
```
jeefberkey: Encourage users to use `assert_type` or other methods

lib/puppet/parser/functions/to_string.rb
```diff
@@ -10,6 +10,8 @@ module Puppet::Parser::Functions
     @return [String]
     EOS
 
+    function_simplib_deprecation(['to_string', 'to_string is deprecated, please use simplib::to_string'])
```
jeefberkey: Encourage users to use `assert_type` or other methods

lib/puppet/parser/functions/validate_port.rb
```diff
@@ -18,6 +18,8 @@ module Puppet::Parser::Functions
     @return [Nil]
     EOS
 
+    function_simplib_deprecation(['validate_port', 'validate_port is deprecated, please use simplib::validate_port'])
```
jeefberkey: Custom type Simplib::Port and friends?

lib/puppet/functions/simplib/validate_port.rb
lnemsick-simp: @jeefberkey There are some modules in which the port is passed in as part of a configuration hash or the port is passed in as a modifier to a string (e.g., '1.2.3.4:80') So, this method is still needed.

lib/puppet/functions/simplib/to_integer.rb
jeefberkey: another example https://github.com/simp/pupmod-simp-iptables/blob/master/manifests/init.pp#L135-L165

lib/puppet/functions/simplib/validate_port.rb
trevor-vaughan: @lnemsick-simp I would like to keep this function, but I think that it can be changed to a pure Puppet function.

```puppet
$foo = 2

unless $foo =~ String { fail("'foo' should be a String, got '${foo}'") }
```

trevor-vaughan: @lnemsick-simp Also, as @jeefberkey indicated in his first post, using `assert_type` should also suffice for most of our `validate` functions.

lib/puppet/parser/functions/to_integer.rb
trevor-vaughan: @jeefberkey This converts to an integer, so `assert_type` won't work here.

lib/puppet/parser/functions/to_string.rb
trevor-vaughan: @jeefberkey This converts to a string, so `assert_type` won't work here.

lib/puppet/parser/functions/validate_port.rb
trevor-vaughan: Hopefully, people won't be using these any longer, but it can still be useful if you're munging through Strings and don't want to write the `if/else` logic yourself.

lib/puppet/functions/simplib/to_integer.rb
trevor-vaughan: @jeefberkey We don't want this to fail, we want it to convert something to an integer I don't think that Puppet can cast variables.

PR #110, opened 2017-09-11
(SIMP-3280) Add SHA256-based algorithm option to ip_to_cron/rand_cron

lnemsick-simp:
SIMP-3280 specifies to replace the IP-modulus-based algorithm in rand_cron (and thus ip_to_cron)
with a SHA256-based algorithm. This request was specifically meant to address a deficiency with
the randomness of client puppet agent runs scheduled by pupmod::agent::cron in certain
circumstances. (By default, pupmod::agent::cron uses rand_cron to determine the cron minute
for client's puppet agent run from the client's IP address.)

Since, empirical analysis of both the original and proposed algorithms for sets of servers in which both
the number of IP addresses is larger than the cron range and the IP addresses are assigned linearly,
the IP-modulus algorithm provides a better distribution across minutes in an hour, this PR instead
provides both algorithms and allows the user to select the algorithm appropriate for their site. A
corresponding change will be made to pupmod-simp-pupmod, to expose the new option.

Note that the crc32-based algorithm for non-IP input to rand_cron will still be available, for backward
compatibility.

### DIFF
--- a/lib/puppet/functions/simplib/rand_cron.rb
+++ b/lib/puppet/functions/simplib/rand_cron.rb
@@ -98,22 +98,9 @@ Puppet::Functions.create_function(:'simplib::rand_cron') do
   end
 
   def generate_sha256_number(input_string)
-    require 'ipaddr'
     require 'digest'
 
-    ip_num = nil
-    begin
-     ip_num = IPAddr.new(input_string).to_i
-    rescue IPAddr::Error
-    end
-
-    num = nil
-    if ip_num.nil?
-      num = Digest::SHA256.hexdigest(input_string).hex
-    else
-      num = Digest::SHA256.hexdigest(ip_num.to_s).hex
-    end
-    num
+    Digest::SHA256.hexdigest(input_string).hex
   end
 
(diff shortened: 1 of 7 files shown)


lnemsick-simp:
To make it fully backward compatible, need to use crc32 calculation when ip_mod algorithm is selected but the input is NOT an IP address

lib/puppet/functions/simplib/rand_cron.rb
```diff
@@ -0,0 +1,144 @@
...
+    begin
+     ip_num = IPAddr.new(input_string).to_i
+    rescue IPAddr::Error
+    end
+
+    num = nil
+    if ip_num.nil?
+      num = Digest::SHA256.hexdigest(input_string).hex
+    else
+      num = Digest::SHA256.hexdigest(ip_num.to_s).hex
```
heliocentric: Is it necessary to convert the ip to an integer, then back to a string before calling SHA256? the hashing algorithm should provide the same statistical distribution from "192.168.251.1" to '12351123516', between numbers.

PR #115, opened 2017-10-25
(SIMP-3358) Deprecate & convert Puppet 3 functions part 4

lnemsick-simp:
- Convert a subset of Puppet 3 functions to Puppet 4 and emit a
  simplib deprecation warning when the Puppet 3 versions as called:
  - simplib::nets2ddq() replaces deprecated nets2ddq().
  - simplib::validate_array_member() replaces deprecated
    validate_array_member().
  - simplib::validate_between() replaces deprecated validate_between().
    The new version fails validation, instead of returning false.
    This behavior consistent with both how the method is used by SIMP
    modules and the error behavior of all other simplib validate
    functions.
  - simplib::validate_bool() replaces deprecated validate_bool_simp().
  - simplib::validate_deep_hash replaced validate_deep_hash.
- In simplib::validate_deep_hash, fixed validate_deep_hash bug in
  which unknown keys in the Hash to check were not detected.

SIMP-3358 #comment 5 more functions converted
SIMP-3904 #close
SIMP-3905 #close
SIMP-3906 #close
SIMP-3907 #close
SIMP-3908 #close

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -50,10 +50,19 @@ unit-test-ruby-2.3:
   script:
     - bundle install
     - bundle exec rake spec
-acceptance-test:
+
+acceptance-test-no-fips:
   stage: test
   tags:
     - beaker
   script:
     - bundle install --no-binstubs --path=vendor
-    - bundle exec rake acceptance
+    - bundle exec rake beaker:suites
+
+acceptance-test-fips:
+  stage: test
+  tags:
+    - beaker
+  script:
+    - bundle install --no-binstubs --path=vendor
+    - BEAKER_fips=yes bundle exec rake beaker:suites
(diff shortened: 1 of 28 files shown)


kendall-moore:
I feel like a few of these don't need Puppet 4 functions because the Puppet 4 data types solve the problem for you, so you can get away with just deprecating the old ones. Namely, validate_between, validate_bool (we really shouldn't even be trying to support string booleans at this point), and validate_array_member. Probably best in all of these cases to just use language constructs instead.

kendall-moore:
Expanding on the last comment (and also scope creeping beyond this ticket), we probably shouldn't have any validate_ functions in our code base at all anymore, except to deprecate the old Puppet 3 ones. If you can't validate your input with the native language type, make a new Puppet 4 type and confine what it can be there instead.

nick-markowski:
@kendall-moore I think validate functions unique to simp *and* used in multiple modules should still be defined here, such as nets2ddq, deep_hash, etc.

kendall-moore:
I see your point about the deep_hash as it's truly unique and you would have to make a strange regex/hash data type for every single module that has something like that in it. Per nets2ddq - is that a validation function? I thought that was for actual data manipulation.

nick-markowski:
@kendall-moore nets2ddq was a bad example, you are correct.

PR #116, opened 2017-10-25
(SIMP-3910) Create MAC address type and test

lnemsick-simp:
* Create Simplib::Macaddress custom type
* Add deprecation notice to validate_macaddress() that references
  Simplib::Macaddress as its replacement
* Write unit test

SIMP-3910 #close


types/macaddress.pp
```diff
@@ -0,0 +1,2 @@
+# Matches MAC addresses
+type Simplib::Macaddress = Pattern['^[0-9a-fA-F]{1,2}:[0-9a-fA-F]{1,2}:[0-9a-fA-F]{1,2}:[0-9a-fA-F]{1,2}:[0-9a-fA-F]{1,2}:[0-9a-fA-F]{1,2}$']
```
jeefberkey: Couldn't each group only be matched twice, never once?

lnemsick-simp: Match of once allows for implied 0.  For example 'f' <-> '0f'.
I

PR #127, opened 2017-11-15
(SIMP-4067) simplib: Disable simplib deprecation warnings by default

lnemsick-simp:
Disable simplib deprecation warnings, temporarily.
Will re-enable when we are ready to address the deprecations
in SIMP modules.

Removed trailing whitespace.

SIMP-4067 #close

### DIFF
--- a/lib/puppet/functions/simplib/deprecation.rb
+++ b/lib/puppet/functions/simplib/deprecation.rb
@@ -1,8 +1,8 @@
 # Function to print deprecation warnings, logging a warning once
 # for a given key.
 #
-# Messages can be suppressed if the SIMPLIB_LOG_DEPRECATIONS
-# environment is set to 'false'
+# Messages can be enabled if the SIMPLIB_LOG_DEPRECATIONS
+# environment variable is set to 'true'
 #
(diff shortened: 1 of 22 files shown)


lib/puppet/functions/simplib/deprecation.rb
```diff
@@ -1,8 +1,8 @@
 # Function to print deprecation warnings, logging a warning once
 # for a given key.
 #
-# Messages can be suppressed if the SIMPLIB_LOG_DEPRECATIONS
-# environment is set to 'false'
+# Messages can be enabled if the SIMPLIB_LOG_DEPRECATIONS
```
jeefberkey: Is this going to be switched back in the future?

lnemsick-simp: @jeefberkey The intent is to switch it back when we are ready to fix the use of deprecated functions in the SIMP modules.  Right now, we are not done converting Puppet 3 functions to Puppet 4 and have, for the most part, have not converted modules using the old functions to use the new functions.

lnemsick-simp: SIMP-4071

PR #129, opened 2017-12-08
(SIMP-4156) Add Simplib::Domain types

op-ct:
This patch introduces Puppet data types that validate DNS domains against
the TLD restrictions from RFC 3696, Section 2.

SIMP-3313 #comment Added DNS domain validation to pupmod-simp-simplib
SIMP-4156 #close


op-ct:
This PR should be merged after #128.

PR #130, opened 2017-12-13
(SIMP-3892) Add login_defs custom fact

trevor-vaughan:
Returns a structured fact for the entire contents of /etc/login.defs
and defaults to returning 'uid_min' and 'gid_min' if they are not
present in the file

SIMP-3892 #comment add login_defs structured fact for use by other modules

### DIFF
--- a/lib/facter/login_defs.rb
+++ b/lib/facter/login_defs.rb
@@ -6,18 +6,8 @@ Facter.add('login_defs') do
   end
 
   setcode do
-    def read_login_defs
-      File.read('/etc/login.defs')
-    end
 
-    if ['RedHat','CentOS'].include?(Facter.value(:operatingsystem)) &&
-       Facter.value(:operatingsystemmajrelease) < '7'
-        id_min = 500
-    else
-      id_min = 1000
-    end
-
-    attribute_hash = read_login_defs.lines.
+    attribute_hash = File.read('/etc/login.defs').lines.
       delete_if{|x| x =~ /^\s*(#|$)/}.
       map{|x| x = x.split(/\s+/); x[0].downcase!; x}.
       to_h
@@ -35,12 +25,6 @@ Facter.add('login_defs') do
         attribute_hash[k] = false
       elsif v =~ /^\d+$/
         attribute_hash[k] = v.to_i
-      end
-    end
-
-    ['uid_min', 'gid_min'].each do |id_attr|
-      unless attribute_hash[id_attr]
-        attribute_hash[id_attr] = id_min
       end
     end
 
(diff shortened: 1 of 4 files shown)


spec/unit/facter/login_defs_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+MD5_CRYPT_ENAB  no
+      EOM
+    }
+
+    it 'should return hash of values from /etc/login.defs with appropriate conversions' do
+      File.expects(:exist?).with('/etc/login.defs').returns(true)
+      File.expects(:readable?).with('/etc/login.defs').returns(true)
+      File.expects(:read).with('/etc/login.defs').returns(login_defs_content)
+
+      File.stubs(:exist?).with(Not(equals('/etc/login.defs')))
```
lnemsick-simp: Given how hard this was to figure out how to mock the File.* commands with mocha when these commands are called in parts of the code that is not under test with different arguments, seems like a comment would be warranted.

spec/unit/facter/login_defs_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+
+  before(:each) do
+    Facter.clear
+
+    Facter.stubs(:value).with(:operatingsystem).returns('Linux')
+  end
+
+  context 'with a well formed /etc/login.defs' do
+    let(:login_defs_content) { <<-EOM
+MAIL_DIR        /var/spool/mail
```
lnemsick-simp: Please add some whitespace lines, comment lines, and empty lines to validate these lines are ignored.

lib/facter/login_defs.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  setcode do
+    def read_login_defs
+      File.read('/etc/login.defs')
+    end
+
+    if ['RedHat','CentOS'].include?(Facter.value(:operatingsystem)) &&
+       Facter.value(:operatingsystemmajrelease) < '7'
+        id_min = 500
```
lnemsick-simp: - Why is this different?  A comment would be helpful.
- This line is untested.

jeefberkey: Can you cite where you got this number from? I don't think there should be hardcoded defaults in facts.

trevor-vaughan: @jeefberkey I've been using EL for a long time :-D.

Ref https://github.com/sindresorhus/default-uid/issues/1#issuecomment-69051381

Basically, 🤷‍♂️

jeefberkey: Oh I meant if you left it in there, you should document where it comes from. Embedding knowledge, etc etc

PR #131, opened 2018-01-04
(SIMP-4214) simplib: Create ipa fact

lnemsick-simp:
Created ipa fact that provides the IPA domain and server to which
a host is connected, when the host is joined to the IPA domain
AND is able to communicate with the IPA server.

As written, this fact cannot to distinguish between the following cases:

* No connectivity to the IPA server when joined to an IPA domain.
* The host is no longer joined to an IPA domain.

Addressing this deficiency is not trivial and will need to be done
in a follow-on ticket.

NOTE: At first blush, it seems like the fact could have been written
to pull information locally from the IPA configuration file,
/etc/ipa/default.conf.  However, the information in this file is not
definitive. Specifically, if the client has been removed from the IPA
domain using the IPA server GUI instead of running the IPA client
uninstall, this file will still exist but be inapplicable.

SIMP-4214 #close
SIMP-4215 #close
SIMP-4162 #comment code, unit, and acceptance test

### DIFF
--- a/spec/acceptance/suites/ipa_fact/ipa_fact_spec.rb
+++ b/spec/acceptance/suites/ipa_fact/ipa_fact_spec.rb
@@ -1,0 +1,71 @@
+require 'spec_helper_acceptance'
+
+test_name 'ipa fact'
+
+describe 'ipa fact' do
+  let (:manifest) {
+    <<-EOS
+      $ipa_value = $facts['ipa']
+      simplib::inspect('ipa_value', 'oneline_json')
+    EOS
+  }
+
+  servers = hosts_with_role(hosts, 'server')
+  servers.each do |server|
+    context 'when IPA is not installed' do
+      it 'ipa fact should be nil' do
+        results = apply_manifest_on(server, manifest)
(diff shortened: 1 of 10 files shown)


lnemsick-simp:
Per discussion with @trevor-vaughan, need to add values from /etc/ipa/default.conf as default values.

spec/acceptance/suites/ipa_fact/ipa_fact_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+
+      it 'ipa fact should have unknown status when connection to IPA server is down' do
+        # stop IPA server
+        on(server, 'ipactl stop')
+
+        fqdn = on(server,'facter fqdn').output.strip
+        results = apply_manifest_on(server, manifest)
+        expect(results.output).to match(/Notice: Type => Hash Content => {"default_domain":"#{ipa_domain}","default_server":"#{fqdn}","status":"unknown","domain":"","server":""}/)
+        results = on(server, 'puppet facts')
+        expect(results.output).to match(/"ipa": {/)
```
jeefberkey: You can totally parse this with JSON.load

PR #137, opened 2018-03-25
(SIMP-4009) Updated for Puppet 5 and OEL

trevor-vaughan:
SIMP-4009 #comment updated simplib


lamawithonel:
The `ipa_fact` suite does not work in FIPS mode, so I've excluded those tests from the `.gitlab-ci.yml` to save time and resources.  Any thoughts on including them, but setting them to `allow_failure`?

lamawithonel:
:heavy_check_mark: https://gitlab.com/simp/pupmod-simp-simplib/pipelines/22516825

Dangerfile
```diff
@@ -0,0 +1,54 @@
...
+end
+
+# Don't allow Work in Progress PRs to merge
+fail('PR is classed as Work in Progress') if github.pr_title.start_with? 'WIP:'
+
+# Reviewers are most effective at evaluating small change-sets as measured
+# in lines of code (LOC).  Less than 50 LOC is ideal, but 200 is more realistic
+# and still within acceptable limits.  After about 400 LOC reviewers tend
+# to skim over most of the code, effectively nullifying their review.  To
+# avoid this, disallow trivial change-sets over 200 LOC, and otherwise warn.
```
op-ct: There's a sort of irony in the LOC this explanatory comment block takes. :sunglasses:

Dangerfile
```diff
@@ -0,0 +1,54 @@
...
+#
+# See: http://support.smartbear.com/support/media/resources/cc/book/code-review-cisco-case-study.pdf#page=16
```
op-ct: * This link covers everything the rest of the comment block says
* The wording of the `warn` statements below cover ~90% of it, too

Dangerfile
```diff
@@ -0,0 +1,54 @@
...
+#
+# TODO: Change warnings to failures once culture has adjusted.  Failure
```
op-ct: Prescriptive cultural TODOs seem like extra LOC for something you won't need to be reminded about if the culture has actually adjusted.

op-ct:
@lamawithonel I agree, leaving out the impossible IPA+FIPS jobs was the right decision.

.gitlab-ci.yml
```diff
@@ -1,259 +1,361 @@
...
-pup4_7-unit:
-  stage: unit
-  tags:
-    - docker
-  image: ruby:2.1
+# ----------------------------------------------------------------------
+# Version Matrix
+#
+# Selections are based on the component version combinations included with
+# currently supported releases of SIMP and Puppet Enterprise.
```
op-ct: Slimming down this section from the [first draft][1] was a step in the right direction.

However, the original comment is still more effective in one simple line:

```bash
# The testing matrix considers ruby/puppet versions supported by SIMP and PE`
```
Than the new comment is with four:

```bash
# Version Matrix 
#
# Selections are based on the component version combinations included with
# currently supported releases of SIMP and Puppet Enterprise.
```
The original version was concise and clear.  The new version is wordier and ambiguous in comparison: 
* The passive voice adds words and makes the point less direct.
* The well-known acronym "PE" is unnecessarily expanded into full words.
* The generic word "Selections" replaced the specific  term "testing matrix."
* As a consequence, two additional lines (starting with "`# Version Matrix`") were needed to reinstate the information that "Selections" had removed.

[1]: https://github.com/simp/pupmod-simp-simplib/blob/03c597f504bd460989d73b0160c3935e2522487f/.gitlab-ci.yml#L86-L93

.gitlab-ci.yml
```diff
@@ -1,259 +1,361 @@
...
+.pe_2016_4: &pe_2016_4
   variables:
     PUPPET_VERSION: '~> 4.7.0'
-  <<: *cache_bundler
-  <<: *setup_bundler_env
-  <<: *spec_tests
 
+.simp_6_0: &simp_6_0
+  variables:
+    PUPPET_VERSION: '~> 4.8.1'
```
op-ct: These `$PUPPET_VARIABLE` declarations were much better off when they were defined in the jobs.   It was slightly redundant, but much clearer to read and maintain.  

Abstracting them into YAML merge keys takes twice as many lines to do the same thing, and is significantly more oblique.

The merge keys are misleadingly named after product releases (e.g., `simp_6_0`, `pe_2016_4`):
  * Their only function is to provide the `$PUPPET_VERSION` variable,
  * All jobs that use these keys are based on the Puppet version (e.g., `el-pup_4_8-default`, `el-pup_4_8-fips`)
  * Multiple SIMP / PE products can require the same puppet version.

.gitlab-ci.yml
```diff
@@ -1,259 +1,361 @@
...
+# Acceptance tests
+#
+
+# Because acceptance tests are so much more expensive than other tests, this
+# test matrix is even more limited.  Here we stick to versions supported
+# by non-EOL meta-releases of SIMP and LTS versions of Puppet Enterprise.
+
+# ----------------------------------------------------------------------
+# Puppet 4.8 for SIMP 6.0 and PE 2016.5
+el-pup4_8-default:
```
op-ct: I suggest putting the version of puppet ahead at the front of job name.  There is limited space on the pipeline UI, and the puppet version is the most significant change.

The `el` is unnecessary; all of these tests are `el`

.gitlab-ci.yml
```diff
@@ -1,259 +1,367 @@
...
-    PUPPET_VERSION: '~> 5.3.0'
-  <<: *cache_bundler
-  <<: *setup_bundler_env
-  <<: *spec_tests
-  allow_failure: true
+# Execute simple sanity checks on the environment before proceeding to more
+# resource-intensive jobs.  Besides running checks, this condenses the initial
+# cache generation into a single job for the later stages.  The first stage,
+# in particular, would otherwise suffer a heavy cache-miss penalty as its
+# jobs kick off in parallel.
```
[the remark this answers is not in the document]
lamawithonel: I've updated it to use the `CI_PROJECT_PATH_SLUG` so it persists across PRs for the entire repo.  I didn't want to use the `CI_NAMESPACE_*` (group-level) variables, because there may be legitimate reasons to use different cache content from repo to repo.  It might be worth sharing those across repo types, i.e., pupmod and rubygem, but I thought that should/could wait.  Also, the upload and download of the cache file takes a lot of time, so I wanted to prevent it from growing forever and ever.  That, too, could use some work outside of this PR, maybe in how the runners are configured, but for now I felt this was the most expedient way to ensure the cache expired.

spec/acceptance/nodesets/default.yml
```diff
@@ -1,20 +1,53 @@
...
+    platform: el-7-x86_64
+    box: centos/7
     hypervisor: vagrant
-  el6:
+    yum_repos:
+      epel:
+        mirrorlist: 'https://mirrors.fedoraproject.org/metalink?repo=epel-7&arch=$basearch'
+        gpgkeys:
+          - https://getfedora.org/static/352C64E5.txt
+  oel6:
```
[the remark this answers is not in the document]
lamawithonel: Oh, no. You're right.

I was thinking the acceptance tests were wrapped in `hosts.each` or similar-- the usual Beaker pattern. If that were the case, each test would execute on each node in the node group.  But that is not the case for this module.

The tests could be wrapped in the appropriate loops, but some of that may conflict with existing loops.  The other option is to duplicate each job for Oracle Linux.  Thoughts?

ayohrling:
acceptance: https://gitlab.com/simp/pupmod-simp-simplib/pipelines/26040570

ayohrling:
acceptance: https://gitlab.com/simp/pupmod-simp-simplib/pipelines/26091395

ayohrling:
acceptance: https://gitlab.com/simp/pupmod-simp-simplib/pipelines/26126051

PR #138, opened 2018-04-16
(SIMP-4798) Improved reboot_notify error handling

trevor-vaughan:
SIMP-4799 #close

### DIFF
--- a/lib/puppet/provider/reboot_notify/notify.rb
+++ b/lib/puppet/provider/reboot_notify/notify.rb
@@ -11,7 +11,7 @@ Puppet::Type.type(:reboot_notify).provide(:notify) do
     @records = {}
 
     begin
-      @records = PSON.parse(File.read(@target))
+      @records = JSON.parse(File.read(@target))
     rescue
       return false
     end
@@ -20,11 +20,15 @@ Puppet::Type.type(:reboot_notify).provide(:notify) do
   end
 
   def create
-    File.open(@target,'w'){|fh| fh.puts(PSON.pretty_generate({}))}
+    begin
+      File.open(@target,'w'){|fh| fh.puts(JSON.pretty_generate({}))}
+    rescue
+      raise(Puppet::Error, "reboot_notify: Could not create '#{@target}'")
+    end
   end
 
   def destroy
-    File.unlink(@target)
+    File.unlink(@target) if File.exist?(@target)
   end
 
   def update
@@ -36,7 +40,11 @@ Puppet::Type.type(:reboot_notify).provide(:notify) do
       :updated => Time.now.tv_sec
     }
 
-    File.open(@target,'w'){|fh| fh.puts(PSON.pretty_generate(@records))}
+    begin
+      File.open(@target,'w'){|fh| fh.puts(JSON.pretty_generate(@records))}
+    rescue
+      raise(Puppet::Error, "reboot_notify: Could not update '#{@target}'")
+    end
   end
 
   def self.post_resource_eval
@@ -45,9 +53,9 @@ Puppet::Type.type(:reboot_notify).provide(:notify) do
     target = "#{Puppet[:vardir]}/reboot_notifications.json"
     records = {}
     begin
-      records = PSON.parse(File.read(target))
+      records = JSON.parse(File.read(target))
     rescue
-      Puppet.error("Could not parse file: #{target}")
+      raise(Puppet::Error, "Could not parse file: #{target}")
     end
 
     current_time = Time.now.tv_sec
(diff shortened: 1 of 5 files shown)


lib/puppet/provider/reboot_notify/notify.rb
```diff
@@ -36,7 +40,11 @@ def update
       :updated => Time.now.tv_sec
     }
 
-    File.open(@target,'w'){|fh| fh.puts(PSON.pretty_generate(@records))}
+    begin
+      File.open(@target,'w'){|fh| fh.puts(JSON.pretty_generate(@records))}
+    rescue
```
lnemsick-simp: See comment above about rescuing specific exceptions and including the specifics of that error in the error message.

lib/puppet/provider/reboot_notify/notify.rb
```diff
@@ -57,18 +65,22 @@ def self.post_resource_eval
...
         msg << ["  #{k} => #{v['reason']}"]
       end
 
       Puppet.notice(msg.join("\n"))
     end
 
-    File.open(target,'w'){|fh| fh.puts(PSON.pretty_generate(records))}
+    begin
+      File.open(target,'w'){|fh| fh.puts(JSON.pretty_generate(records))}
+    rescue
```
lnemsick-simp: See comment above about rescuing specific exceptions and including the specifics of that error in the error message.

lib/puppet/provider/reboot_notify/notify.rb
```diff
@@ -20,11 +20,15 @@ def exists?
   end
 
   def create
-    File.open(@target,'w'){|fh| fh.puts(PSON.pretty_generate({}))}
+    begin
+      File.open(@target,'w'){|fh| fh.puts(JSON.pretty_generate({}))}
+    rescue
```
lnemsick-simp: I'm not a fan of blanket rescues and would prefer to see rescue of specific error types or their base class, and a reporting of the error they provide.  For example,

rescue Errno::ENOENT, Errno::EACCES, JSON::JSONError, ArgumentError, TypeError => e
  *OR*
rescue SystemCallError, JSON::JSONError => e
  raise(Puppet::Error, "reboot_notify:  Could not create '#{@target}' - #{e.message}")

lib/puppet/provider/reboot_notify/notify.rb
```diff
@@ -11,7 +11,7 @@ def exists?
     @records = {}
 
     begin
-      @records = PSON.parse(File.read(@target))
+      @records = JSON.parse(File.read(@target))
     rescue
```
lnemsick-simp: Would rather see rescue of exceptions raised by JSON.parse() and File.read(), instead of blanket exception.  For example,

raise JSON::JSONError,ArgumentError,TypeError,SystemCallError

lib/puppet/provider/reboot_notify/notify.rb
```diff
@@ -45,9 +53,9 @@ def self.post_resource_eval
     target = "#{Puppet[:vardir]}/reboot_notifications.json"
     records = {}
     begin
-      records = PSON.parse(File.read(target))
+      records = JSON.parse(File.read(target))
     rescue
```
lnemsick-simp: Would rather see rescue of exceptions raised by File.read() and JSON.parse() and reporting of the error these exceptions provide. . For example,

raise JSON::JSONError,ArgumentError,TypeError,SystemCallError => e
   raise(Puppet::Error, "Could not parse file: #{target} - #{e.message}")

lib/puppet/provider/reboot_notify/notify.rb
```diff
@@ -57,18 +65,22 @@ def self.post_resource_eval
...
     }
 
-    if not records.empty? then
+    unless records.empty?
       msg = ["System Reboot Required Because:"]
 
       records.each_pair do |k,v|
         # This is a fail safe for empty 'reasons'
-        records[k]['reason'] = 'modified' if ( records[k]['reason'].nil? or records[k]['reason'].empty? )
+        records[k]['reason'] = 'modified' if ( records[k]['reason'].nil? || records[k]['reason'].empty? )
```
lnemsick-simp: Why the change from 'or' to '||'?  They are equivalent and the first way is more understandable.

spec/unit/puppet/provider/reboot_notify/notify_spec.rb
```diff
@@ -0,0 +1,170 @@
+require 'spec_helper'
```
lnemsick-simp: Excellent test

lib/puppet/provider/reboot_notify/notify.rb
trevor-vaughan: Consistence with the rest of the codebase. We ended up moving to `||` from `or` across the board in the past in theory due to execution prioritization ordering.

PR #141, opened 2018-05-03
(SIMP-4885) Add hash_to_opts function

jeefberkey:
- Add `simplib::hash_to_opts` which turns a hash into a string with each
  key prefixed with '--' and connected to each value with '='. Useful
  for generating commands

SIMP-4885 #close

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,68 +1,259 @@
-# 
+# The testing matrix considers ruby/puppet versions supported by SIMP and PE:
+#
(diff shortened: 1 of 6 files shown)


jeefberkey:
:clock4: https://gitlab.com/simp/pupmod-simp-simplib/pipelines/21482609

jeefberkey:
> Short => Sort option -q and joined to the value with a space

What do you mean by this? 

I agree with the first point.

trevor-vaughan:
@jeefberkey The Short comment is due to the following:

`--foo=bar` is usually equivalent to `-f bar`

jeefberkey:
@trevor-vaughan that case is covered now, and tested.

PR #142, opened 2018-05-29
(SIMP-4915) Add array_style to hash_to_opts

jeefberkey:


functions/hash_to_opts.pp
```diff
@@ -25,18 +27,25 @@ function simplib::hash_to_opts (
   Struct[{
     Optional[connector]       => String[1],
     Optional[prefix]          => String[1],
+    Optional[array_style]     => Enum['comma','repeat'],
```
trevor-vaughan: I would change this to `delimiter` and `repeat`.

jeefberkey: Those are better names, I like them

PR #146, opened 2018-06-19
(SIMP-4654) Fix runlevel enforcement

trevor-vaughan:
The `level` property functions in the providers were never actually
hooked back into the type. This meant that systems wouldn't actually
ever properly switch runlevels when using the `runelvel` type.

SIMP-4654 #comment Fix runlevel native type


lib/puppet/functions/simplib/module_exist.rb
```diff
@@ -0,0 +1,18 @@
+# Determines if a module exists in the current environment
```
lnemsick-simp: why was this added to this pr?

lib/puppet/type/runlevel.rb
```diff
@@ -29,16 +29,35 @@ def initialize(*args)
     end
   end
```
lnemsick-simp: where is the test for these changes

lib/puppet/functions/simplib/module_exist.rb
trevor-vaughan: It's also for SIMP-4654

lib/puppet/type/runlevel.rb
trevor-vaughan: Right here: https://github.com/simp/pupmod-simp-gdm/pull/34/files#diff-9253f3c9cbffa1500f41d4db2baa8664R29

jeefberkey:
:clock4: https://gitlab.com/simp/pupmod-simp-simplib/pipelines/24403827

PR #148, opened 2018-07-01
(SIMP-4447) Add 'assert_optional_dependency' func

trevor-vaughan:
Added a function `assert_optional_dependency` that allows users to fail
if expected functionality is not present in the current environment's
module set. This provides the ability for users to support multiple
vendor modules without forking.

SIMP-4447 #close

### DIFF
--- a/lib/puppet/functions/simplib/assert_optional_dependency.rb
+++ b/lib/puppet/functions/simplib/assert_optional_dependency.rb
@@ -1,0 +1,170 @@
+# Fails a compile if the system does not contain a correct version of the
+# required module in the current environment.
+#
+# Provides a message about exactly which version of the module is required.
+#
+Puppet::Functions.create_function(:'simplib::assert_optional_dependency') do
+  # @param source_module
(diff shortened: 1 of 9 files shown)


lib/puppet/functions/simplib/assert_optional_dependency.rb
```diff
@@ -0,0 +1,170 @@
...
+    require 'semantic_puppet'
+
+    _module_author, _module_name = module_name.split('/')
+
+    unless _module_name
+      _module_name = _module_author.dup
+      _module_author = nil
+    end
+
+    unless call_function('module_exist', _module_name)
```
nick-markowski: should be simplib::module_exist

nick-markowski:
Currently, if you assert an optional dependency on a module which does not exist in the source module's metadata.json, there is no warning.  I believe there should be a warning; something to the effect of: '"module_a" has unsatisfied dependency "module_b" in environment "production" at /wherever/this/is/asserted"

PR #149, opened 2018-07-27
(SIMP-5096) Fixes to the runlevel type

trevor-vaughan:
* Added a configurable timeout for changing runlevels based on issues
  discovered in the field with systemctl.
* Fixed bugs in the EL6 runlevel persistence where, in some cases, the
  runlevel line might not get written to /etc/inittab.

SIMP-5096 #close

### DIFF
--- a/lib/puppet/provider/runlevel/systemd.rb
+++ b/lib/puppet/provider/runlevel/systemd.rb
@@ -11,11 +11,19 @@ Puppet::Type.type(:runlevel).provide(:systemd) do
   end
 
   def level=(should)
-    return execute([command(:systemctl),'isolate',init2systemd(@resource[:name])])
+    require 'timeout'
+
+    begin
+      Timeout::timeout(@resource[:transition_timeout]) do
+        execute([command(:systemctl),'isolate',init2systemd(@resource[:name])])
+      end
+    rescue Timeout::Error
+      raise(Puppet::Error, "Could not transition to runlevel #{@resource[:name]} within #{@resource[:transition_timeout]} seconds")
+    end
   end
 
   def persist
-    if execute([command(:systemctl),'get-default']).strip == init2systemd(@resource[:name]) then
+    if execute([command(:systemctl),'get-default']).strip == init2systemd(@resource[:name])
       return :true
     else
       return :false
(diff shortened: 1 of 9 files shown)


lib/puppet/provider/runlevel/telinit.rb
```diff
@@ -37,17 +45,29 @@ def persist=(should)
...
       end
     end
+
+    unless found_line
+      newfile << "id:#{@resource[:name]}:initdefault:nil\n"
+    end
+
     inittab.close
 
+    fh = IO.open(IO.sysopen('/tmp/foo','w'))
```
lnemsick-simp: Is this debug that should be removed?

trevor-vaughan: @lnemsick-simp Yep

lib/puppet/provider/runlevel/systemd.rb
```diff
@@ -11,11 +11,19 @@ def level
...
   def level=(should)
-    return execute([command(:systemctl),'isolate',init2systemd(@resource[:name])])
+    require 'timeout'
+
+    begin
+      Timeout::timeout(@resource[:transition_timeout]) do
+        execute([command(:systemctl),'isolate',init2systemd(@resource[:name])])
+      end
+    rescue Timeout::Error
+      raise(Puppet::Error, "Could not transition to runlevel #{@resource[:name]} within #{@resource[:transition_timeout]} seconds")
```
lnemsick-simp: How do you know the command is terminated with this raise?

trevor-vaughan:
@lnemsick-simp Investigate their system and figure out what's going on. I don't know why exactly this is happening, just that it is so I don't know what to tell people. We just fail saying that we couldn't switch runlevels and the user needs to investigate their system.

While I found that switching to runlevel 3 and back worked on some systems, it did not work on all of them and some were just irrecoverable even after an update.

This seems to be a bug in `systemd` but I don't know exactly how to reproduce it yet.

lib/puppet/provider/runlevel/systemd.rb
op-ct: :arrow_up: This is a good question.  

Shouldn't we explicitly test for/kill it in this `rescue` block before raising the `Puppet::Error`?

lib/puppet/provider/runlevel/systemd.rb
```diff
@@ -11,11 +11,19 @@ def level
   end
 
   def level=(should)
-    return execute([command(:systemctl),'isolate',init2systemd(@resource[:name])])
+    require 'timeout'
+
+    begin
+      Timeout::timeout(@resource[:transition_timeout]) do
+        execute([command(:systemctl),'isolate',init2systemd(@resource[:name])])
```
lnemsick-simp: Please add logic that ensures the transition is not already in progress.

lnemsick-simp:
Tested this logic on a workstation that I can reliably can get into the ``systemctl isolate graphical`` hung state.  Operated as expected.

PR #151, opened 2018-08-31
(SIMP-MAINT) Fix spec tests on Puppet 5

jeefberkey:
A new warning has been issued on Puppet 5.5.6, causing the `simplib::deprecation` tests to catch unexpected output.


jeefberkey:
:heavy_check_mark: https://gitlab.com/simp/pupmod-simp-simplib/pipelines/29090342

PR #153, opened 2018-09-27
(SIMP-5383) Fix bug in `puppet_settings` fact

op-ct:
Before this patch, the `puppet_settings` fact incorrectly interpolated
variables using the `[main]` section's settings instead of `[master]`.

A significant consequence of this has been that the fact supplies
 incorrect paths for `[master]` settings, notably `server_datadir`:

```sh
[root@puppet ~]# puppet config print --section master server_datadir
  /opt/puppetlabs/server/data/puppetserver/server_data

[root@puppet ~]# facter -p puppet_settings.master.server_datadir
  /opt/puppetlabs/puppet/cache/server_data
```

As a consequence, `pupmod::master::sysconfig` ends up using the puppet
agent's vardir instead of puppetserver's (SIMP-5021), causing the
puppetserver service to fail when upgrading to new versions of
puppet-agent (SIMP-5383)―particularly when upgrading from SIMP 6.1 to
SIMP 6.2.

This patch fixes the issue by using `Puppet.override` to temporarily set
Puppet's run_mode to each section while interpolating their settings.

SIMP-5021 #close #comment Fixed bug in `puppet_settings` fact
SIMP-5383 #close #comment Fixed bug in `puppet_settings` fact

### DIFF
--- a/lib/facter/simp_puppet_settings.rb
+++ b/lib/facter/simp_puppet_settings.rb
@@ -3,24 +3,49 @@
 # automatically be able to use the various aspects of the puppet system
 # regardless of the node's platform.
 #
-# Each entry is under a hash entry of its associated section
+# Each entry is under a hash entry of its associated section.
 
 Facter.add(:puppet_settings) do
-  setcode do
-    retval = {}
+   if Object.const_defined?('Puppet') && Puppet.respond_to?(:settings)
+    setcode do
+      retval = {}
+      puppet_settings = Hash[Puppet.settings.map { |k, v| [k, v] }]
 
-    if Object.const_defined?('Puppet') && Puppet.respond_to?(:settings)
-      puppet_settings = Hash[Puppet.settings.map{|k,v| [k,v]}]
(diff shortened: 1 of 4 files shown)


lib/facter/simp_puppet_settings.rb
```diff
@@ -3,24 +3,49 @@
...
-  setcode do
-    retval = {}
+   if Object.const_defined?('Puppet') && Puppet.respond_to?(:settings)
+    setcode do
+      retval = {}
+      puppet_settings = Hash[Puppet.settings.map { |k, v| [k, v] }]
 
-    if Object.const_defined?('Puppet') && Puppet.respond_to?(:settings)
-      puppet_settings = Hash[Puppet.settings.map{|k,v| [k,v]}]
+      Puppet.settings.eachsection do |section|
```
jeefberkey: Maybe we can get this data from [Puppet-as-a-library](https://puppet.com/docs/puppet/5.4/release_notes.html#puppet-as-a-library-pal-tasks-and-scripting) in the future

PR #154, opened 2018-09-27
Fix errors on non-Linux OS's

jadestorm:
Hi folk, this PR takes care of the following two errors that show up on non-Linux OS's:

Under Windows:
```
puppet : [1;31mError: Facter: error while resolving custom fact "simplib_sysctl": undefined method `strip' for nil:NilClass[0m
    + CategoryInfo          : NotSpecified: ([1;31mError: F...il:NilClass[0m:String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
```
This was due to it not having sysctl at all.

Under Mac OS X (and Windows if I recall correctly):
```
error while resolving custom fact "uid_min": undefined method `empty?' for nil:NilClass
```
This was due to some slightly confused unless login (at least per my and my coworkers read).  We simplified the section a bit.

Let me know if you have any concerns or questions about the patch!

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,10 @@
+* Fri Sep 28 2018 Daniel Henninger <daniel@vorpalcloud.org> - 3.11.1-0
+- Fixed bug where uid_min would throw errors under operating systems
+  without /etc/login.defs.
+- Fixed bug where simplib_sysctl would throw an undefined method error
+  on non-Linux OS's.  (both those with sysctl (MacOS X) and without (Windows))
+- Both patches of which were improved by Trevor Vaughan <tvaughan@onyxpoint.com>.
+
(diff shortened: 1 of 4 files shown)


trevor-vaughan:
@jadestorm Thanks for the patch!

It looks like we should probably just confine these facts to existence of the target artifacts. Would you be up for testing an update if we update this PR?

trevor-vaughan:
@jadestorm I've patched this with `confine` statements. Can you make sure it still works for you?

lnemsick-simp:
@jadestorm Would you please update the CHANGELOG and metadata.json with a patch version bump?

jadestorm:
@trevor-vaughan Hrm, I don't actually know how to get at this edit.  How do I check it out?

jadestorm:
Scratch that -- I didn't realize the update went back to my fork.

jadestorm:
@lnemsick-simp Done!

PR #163, opened 2019-01-04
(SIMP-5921) Consolidate TPM facts in simplib

lnemsick-simp:
- Add has_tpm and tpm facts originally found in the simp/tpm module
- Add tpm2 fact originally found in the simp/tpm2 module
- Add tpm_1_2_enabled and tpm_2_0_enabled facts developed for the
  simp/hirs_provisioner module

SIMP-5921 #close


lib/facter/tpm/util.rb
```diff
@@ -0,0 +1,208 @@
...
+          output['_status'] = 'Trousers is not running'
+          return output
+        else
+          version = YAML.load(cmd_out)
+
+          # Format keys in a way that Facter will like
+          begin
+            output = Hash[version.flatten[1].map{ |k,v| [k.downcase.gsub(/ /, '_'), v] }]
+          rescue NoMethodError
+            version.delete_if { |k,y| k =~ /Version Info/ }
```
sharkbruhaha: should be |k,v| ?

lib/facter/tpm/util.rb
```diff
@@ -0,0 +1,208 @@
...
+      #  ```
+      #
+      def version
+        require 'yaml'
+
+        output = Hash.new
+        cmd_out = tpm_version
+
+        if cmd_out == "" or cmd_out.nil?
+          output['_status'] = 'Trousers is not running'
```
sharkbruhaha: should we be using the normal spelling: TrouSerS ?

lib/facter/tpm/util.rb
```diff
@@ -0,0 +1,208 @@
...
+              out['_status'] = 'success: raw pubek grabbed' unless raw.nil?
+            end
+          else
+            out['_status'] = 'error: password file not found'
+            return out
+          end
+        end
+
+        if raw.nil? or raw == ""
+          out['_status'] = 'error: trousers is not running, the tpm is not enabled, or the password in the password file is incorrect'
```
sharkbruhaha: TrouSerS ?

lib/facter/tpm/util.rb
lnemsick-simp: What you call the variables is entirely up to you in Ruby

lnemsick-simp:
Need to pull in https://github.com/simp/pupmod-simp-tpm2/pull/6 and the utils tests that I failed to port over.

PR #166, opened 2019-01-18
(SIMP-5940) Add simplib::dlookup function

trevor-vaughan:
- Add a 'defined type' lookup capability, `simplib::dlookup` that provides a
  consistent method for retrieving defined type parameters from Hiera in an
  opt-in manner.
- Fixed various YARD documentation issues

SIMP-5940 #comment Add function to support defined type global lookups

### DIFF
--- a/lib/puppet/functions/simplib/dlookup.rb
+++ b/lib/puppet/functions/simplib/dlookup.rb
@@ -12,10 +12,7 @@
 # appropriately but was split out in case the underlying syntax needs to
 # change in the future.
 #
-# There are two ways to call this method and, in both cases, to prevent
-# conflicts with class parameters, the keys in Hiera have been modified to
-# completely avoid items that are allowed Puppet Class names but to also mirror
-# common Puppet nomenclature for working with existing resources.
+# There are two ways to call this method as shown in the following examples.
 #
 # @example Global Options
 #
--- a/lib/puppet/functions/simplib/dlookup.rb
+++ b/lib/puppet/functions/simplib/dlookup.rb
@@ -131,7 +131,7 @@ Puppet::Functions.create_function(:'simplib::dlookup') do
     retval = call_function('simplib::lookup', target_param, options)
 
     # Fall back to the global lookup for this option
-    if options['default_value'] && retval == options['default_value']
+    if retval == options['default_value']
       retval = dlookup(define_id, param, options)
     end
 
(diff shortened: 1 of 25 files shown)


trevor-vaughan:
:hourglass_flowing_sand: https://gitlab.com/simp/pupmod-simp-simplib/pipelines/44199419

lib/puppet/functions/simplib/dlookup.rb
```diff
@@ -0,0 +1,140 @@
...
+#
+# This function introduces the capability to consistently opt-in to a lookup
+# syntax for overriding all parameters of a given defined type or parameters on
+# a specific instance of a defined type.
+#
+# This calls `simplib::lookup` under the hood after formatting the parameters
+# appropriately but was split out in case the underlying syntax needs to
+# change in the future.
+#
+# There are two ways to call this method and, in both cases, to prevent
```
lnemsick-simp: This comment only applies to looking up parameters for a specific defined type.

lib/puppet/functions/simplib/dlookup.rb
```diff
@@ -0,0 +1,140 @@
...
+#   ```
+#
+#   Example Hieradata:
+#
+#   ```yaml
+#   ---
+#   mydef::global::ssl_version: 'TLS1.2'
+#   ```
+#
+# @example Specific Instance Options
```
lnemsick-simp: This is not the API I was envisioning.  This says that when you create the define, you have to specify whether it can be globally overridden or individually overridden .  I was expecting a lookup that would first look for an individual override, then look for a global override and finally look for the value when the define was made.

spec/spec_helper.rb
```diff
@@ -132,11 +155,32 @@ def set_hieradata(hieradata)
...
 
+    hiera_datadir = File.dirname(c.hiera_config)
+
     # sanitize hieradata
     if defined?(hieradata)
-      set_hieradata(hieradata.gsub(':','_'))
+      custom_hieradata = hieradata.gsub(/(:|\s)/,'_')
     elsif defined?(class_name)
-      set_hieradata(class_name.gsub(':','_'))
+      custom_hieradata = class_name.gsub(/(:|\s)/,'_')
```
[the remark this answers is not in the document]
trevor-vaughan: @sharkbruhaha They cannot have spaces.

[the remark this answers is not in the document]
sharkbruhaha: is that not what the 'gsub(/(:|\s)/,'_')' is? 
      a substitution of (colons OR spaces) for underscores

lib/puppet/functions/simplib/dlookup.rb
trevor-vaughan: @lnemsick-simp This is basically just a more user friendly way of achieving the same thing without needing to pass around option hashes and bloat the calls even more in the parameter calls (they're getting pretty darn big).

PR #167, opened 2019-01-28
Added execute timeout

hanej:
Changed Facter::Core::Execution.exec to Facter::Core::Execution.execute and added a timeout so the fact doesn't hang indefinitely.


trevor-vaughan:
@hanej Thanks for the patch! One question, how reasonable is `10` seconds for an IPA timeout? That seems just fine for Kerberos but considering the number of subsystems that IPA handles I was wondering if something like `30` might be better.

Also, the tests are currently failing. I'd be happy to update them if you don't mind me pushing to your branch.

PR #168, opened 2019-01-31
[Don't merge this!]  Purge functions that are already in stdlib

jadestorm:
Hi folk!  So I wanted to submit this PR not to be accepted as-is, but to start a quick discussion and alert you to something I ran into.

I've been running stdlib 5.x for quite some time now (the current latest version) and simplib alongside of it for the gdm/gnome/dconf modules.  To date, there has been no obvious problems.

However, I am testing upgrading to Puppet 6 and I discovered that deep_merge, and validate_integer conflict with stdlib now.  I guess puppet just ignored the fact that they were defined twice before, but now it freaks out:

```
2019-01-31T11:10:16.574-05:00 ERROR [qtp1554091179-74] [p.r.core] Internal Serve
r Error: org.jruby.exceptions.SecurityError: (SecurityError) Illegal method defi
nition of method 'deep_merge' on line 13' in legacy function. See https://puppet
.com/docs/puppet/latest/functions_refactor_legacy.html for more information
```

So in this fork, I simply removed the deep_merge and validate_integer functions that already exist in stdlib and things appear to be working just fine.  Now, currently, simplib does not require stdlib.  However, all of the other simp modules I spot checked did require it, so I don't think it would be hardship for end users to simply nuke these two functions and add stdlib as a requirement to simplib itself.  (note: the other simp modules do actually state stdlib < 5.0.0, and I believe 5.x is when deep_merge showed up)

If you actually want me to make this a formal PR that you could accept, just let me know and I'll update the changelog and readme and such.  However, I mostly wanted to bring the issue up for discussion.


trevor-vaughan:
@jadestorm Thanks for the PR! We were actually just discussing this very issue and realize that we need to get all of our legacy functions namespaced.

I guess since puppet is now raising errors, this makes it a bit more pressing. The big issue that we've had is trying to figure out how to make sure that we don't break existing code for users but it seems like this may be something that we simply need to get out in `6.4.0`.

It's unfortunate, that `stdlib` didn't just pull our functions when they could so that the APIs would match but that's the way it goes sometimes.

jadestorm:
=)  Yeah probably namespacing them will be the best course of action.  You may have noticed that I added another commit after posting that that got rid of the ipaddresses function.  That conflicted with saz/ssh, and I also pinged him about namespacing his.

BTW for reference, you can see the actual change in Puppet Server 6.2.0 that caused me to run into this:

> Added protection against illegal methods in legacy functions
> Puppet now protects against illegal method definitions in loaded legacy functions. Illegal methods in legacy functions disrupt the entire system and can cause difficult-to-diagnose issues. For information on how to remove such methods from legacy functions, see the topic about refactoring legacy functions. PUP-9294

https://puppet.com/docs/puppet/6.2/release_notes_puppet.html#release-notes-puppet

Don't know if that's helpful to y'all or not.  =)

lnemsick-simp:
https://simp-project.atlassian.net/browse/SIMP-6259 and it's corresponding PR https://github.com/simp/pupmod-simp-simplib/pull/174 will address deep_merge

lnemsick-simp:
https://simp-project.atlassian.net/browse/SIMP-6308 and it's corresponding PR https://github.com/simp/pupmod-simp-simplib/pull/176 will address validate_integer

trevor-vaughan:
@jadestorm Can you see if the latest `master` branch takes care of what you need?

We can't deconflict `ipaddresses` at this time and I wonder why a generic SSH module would have a generic SSH fact for ipaddresses so you may want to file a bug about that upstream.

trevor-vaughan:
@jadestorm Closing this but feel free to re-open it with further discussion if necessary!

PR #173, opened 2019-03-08
(SIMP-6182) Creating cron datatype 

sharkbruhaha:
Added custom cron datatype and type_alias unit tests. They Support either a restricted integer as appropriate, or full steps/ranges/names.
Simplib::CronHour
Simplib::CronMinute
Simplib::CronMonthDay
Simplib::CronMonth
Simplib::CronWeekDay

This enables us to now have steps: '*/#' and step ranges: '#-#/#' as well as support for day/month name, and proper numerical bounds on inputs instead of general Integer as has been the trend; with all input verified.


trevor-vaughan:
Also, I think that these should be namespaced under `Cron` so it's `simplib::cron::hour`, etc...

trevor-vaughan:
Also, just took a look at the [cron resource type](https://puppet.com/docs/puppet/5.5/types/cron.html) and this data type doesn't align with what we feed into the Puppet resource from what I can tell.

Conversely, it also doesn't match some of the items in the `crontab(5)` man page.

I think that we should not provide this and should defer it to processing by the resource type even though that is slightly more confusing for the user if they provide incorrect data.

sharkbruhaha:
> 
> 
> Also, I think that these should only be the namespaced version under `Cron` so it's `simplib::cron::hour`, etc...

I do agree it looks nicer all being namespaced under cron, should I move it down as is, or combine the functions into one file per item. I was doing the Cron### after seeing the DomainList as an example of doing similar. Didn't think it would be as clean to do all in one, because it would result in some copy paste regex, unless you can declare multiple functions in the same type file:
`Cron### = Variant[bla bla bla]
hour = Varient[blah blah blah]`

sharkbruhaha:
> 
> 
> Also, just took a look at the [cron resource type](https://puppet.com/docs/puppet/5.5/types/cron.html) and this data type doesn't align with what we feed into the Puppet resource from what I can tell.
> 
> Conversely, it also doesn't match some of the items in the `crontab(5)` man page.
> 
> I think that we should not provide this and should defer it to processing by the resource type even though that is slightly more confusing for the user if they provide incorrect data.

Puppet accepts beyond this even, but I was only including what [cron](https://en.wikipedia.org/wiki/Cron#CRON_expression) allows for. Puppet does translations of both the 3letter shorthand and the fully written our day/month to their numerical counterpart. 

That being said, I am fine with not including matching for the words, as I have never found it to be used that much and the functionality is limited compared to using the numbers.

I do disagree with dropping this though, as our current method of validation is woefully inadequate (and also not even uniform amongst our modules), and is blocking people from actually using steps and ranges which, speaking from personal experience of using older simp was quite annoying and resulted in modifying the modules to just have no verification at all with `String =  ###`, and having to keep up with that on upgrades.

trevor-vaughan:
@sharkbruhaha So the case insensitivity was correct, but you should make that the first match in the set of OR'd regexes so that the least amount of work needs to be done for those options.

sharkbruhaha:
> 
> 
> eye +1 The code looks visually effective but I'd like a second review.
> 
> The README needs to be updated with the new names.

thanks for the catch

sharkbruhaha:
@trevor-vaughan Changed structure back to my original (with `Simplib::Cron::Hour` being the default), and made the base regex type `###_solo`. Having "Array" on the name just felt and looked terrible when I was testing it in rkhunter, and I believe the use cases of people actually wanting to restrict cron input to a single value (which could still be a step/range) is minuscule, so putting the onus on them to have  `Simplib::Cron::Hour_solo` instead seems fine.

PR #174, opened 2019-03-19
(SIMP-6297) Remove `deep_merge()` function

op-ct:
simp-simplib and puppetlabs-stdlib have both provided a Puppet 3.x-style
`deep_merge()` function for a long time, and one of them has to go.

This should be considered a bugfix and not a breaking change—as noted in
SIMP-6259, puppetlabs-stdlib is a dependency of simplib, and its
`deep_merge()` implementation honors the same arguments, provides a
superior implementation (with support for reasonable edge cases), and is
backed up by unit and acceptance tests.

SIMP-6297 #close
SIMP-4942 #comment Generated updated REFERENCE.md + linked README docs
SIMP-6299 #close
SIMP-6300 #close
SIMP-6310 #close


CHANGELOG
```diff
@@ -1,3 +1,10 @@
+* Tue Mar 19 2019 Chris Tessmer <chris.tessmer@onyxpoint.com> - 3.14.1-0
+- Removed simplib's `deep_merge()` 3.x function
+  - NOTE: This is a bugfix and not a breaking change; simplib depends on
+    stdlib, which has long since provided a compatible `deep_merge()` function
+    in the same (global/Puppet 3.x) namespace, but with a superior
+    implementation.  For more details, see SIMP-6259.
```
lnemsick-simp: On isolated networks, I don't think referencing a JIRA ticket is helpful

op-ct: I don't think providing a reference ID is any less helpful than referring to a public URL under the same circumstances.  Both have precedent in our current changelogs.

In this case, providing the JIRA issue provided the most helpful balance I could find between either derailing the focus of the (already large) changelog entry with a deep comparison of implementation details, or just not explaining why removing an entire function did not constitute an API-breaking change.

op-ct:
> simplib now depends upon stdlib, as it uses deep_merge for lib/puppet/provider/reboot_notify/notify.rb. and functions/assert_metadata.pp.
> So, IMHO the metadata.json should be updated to reflect this

Boy, I'd say it should—simplib has _always_ relied on stdlib: at a glance, it's currently using `ensure_packages()`, `load_module_metadata()`, `pick()`, `prefix()`, `delete()`, and the class `stdlib::stages` (Even the outgoing implementation of `deep_merge()` used stdlib's `merge()`)!

I have no idea why `stdlib` is _not_ currently a formal dependency, but I'll add it right now.

PR #180, opened 2019-03-21
(SIMP-6320) Deprecate duplicate Puppet 3 functions

lnemsick-simp:
Deprecated simplib Puppet 3 functions for which corresponding,
though not identically behaving, Puppet builtins or stdlib
functions exist.  The functions will be removed in a future
release, unless SIMP receives requests from users to convert
them to namespaced Puppet 4.x functions.
- array_include() can often be replaced with stdlib member()
- array_size() can often be replaced with the Puppet builtin length()

SIMP-6320 #close

### DIFF
--- a/lib/puppet/parser/functions/array_include.rb
+++ b/lib/puppet/parser/functions/array_include.rb
@@ -21,6 +21,8 @@ module Puppet::Parser::Functions
 
     ENDHEREDOC
 
+    function_simplib_deprecation(['array_include', "array_include is deprecated, please use stdlib's member"])
+
     unless args.length == 2
       raise Puppet::ParseError, ("array_include(): wrong number of arguments (#{args.length}; must be 2)")
     end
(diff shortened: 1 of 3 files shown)


lib/puppet/parser/functions/array_include.rb
```diff
@@ -21,6 +21,8 @@ module Puppet::Parser::Functions
 
     ENDHEREDOC
 
+    function_simplib_deprecation(['array_include', "array_include is deprecated, please use stdlib's member"])
```
trevor-vaughan:
```suggestion
    function_simplib_deprecation(['array_include', "array_include is deprecated, please use puppetlabs-stdlib's 'member()'"])
```

lib/puppet/parser/functions/array_size.rb
```diff
@@ -9,6 +9,8 @@ module Puppet::Parser::Functions
     @return [Integer]
     ENDHEREDOC
 
+    function_simplib_deprecation(['array_size', "array_size is deprecated, please use builtin length"])
```
trevor-vaughan:
```suggestion
    function_simplib_deprecation(['array_size', "array_size is deprecated, please use builtin 'length()'"])
```

PR #182, opened 2019-03-22
(SIMP-3933) Update bracketize, deprecate old one

sharkbruhaha:


lib/puppet/parser/functions/bracketize.rb
```diff
@@ -8,6 +8,8 @@ module Puppet::Parser::Functions
     @return [Variant[String, Array[String]]]
     ENDHEREDOC
 
+function_simplib_deprecation(['bracketize', 'bracketize is deprecated, please use tcpwrappers::bracketize'])
```
trevor-vaughan:
```suggestion
function_simplib_deprecation(['bracketize', 'bracketize is deprecated'])
```

sharkbruhaha: thanks for that catch, forgot to go back to that after one of our talks.

CHANGELOG
```diff
@@ -1,3 +1,7 @@
+* Fri Mar 22 2019 Joseph Sharkey <shark.bruhaha@gmail.com> - 3.15.0-0
+- Marked bracketize puppet 3 function as deprecated.
+- Created bracketize function, and spec test
```
lnemsick-simp: This comment appears to be incorrect.  I don't see any namespaced 4.x bracketize functions in this PR.

CHANGELOG
```diff
@@ -1,3 +1,7 @@
+* Fri Mar 22 2019 Joseph Sharkey <shark.bruhaha@gmail.com> - 3.15.0-0
+- Marked bracketize puppet 3 function as deprecated.
```
lnemsick-simp: Need to note:  "This function will be removed in a future release, unless SIMP receives a request from users to convert it to a namespaced Puppet 4.x function'.
This matches other function deprecation CHANGELOG entries.

sharkbruhaha: It is converted to a puppet 4 function...

sharkbruhaha: *DOH!* I forgot to add the files...

Fixing in push soon

lnemsick-simp: Function and its test were not added to git, so not visible in the PR

lib/puppet/functions/simplib/bracketize.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  def bracketize_string_input(ipaddr_string)
+    ipaddr = ipaddr_string.split(/\s|,|;/).delete_if{ |y| y.empty? }
+    bracketize(ipaddr)
+  end
+
+  def bracketize(ipaddr)
+    require 'ipaddr'
+    ipaddr = Array(ipaddr).flatten
```
lnemsick-simp: Input is not supposed to be modified.  Need to assign to a different (local) variable

lib/puppet/functions/simplib/bracketize.rb
```diff
@@ -0,0 +1,63 @@
...
+  dispatch :bracketize do
+    required_param 'Array[String]', :ipaddr
+  end
+
+  dispatch :bracketize_string_input do
+    required_param 'String', :ipaddr_string
+  end
+
+  def bracketize_string_input(ipaddr_string)
+    ipaddr = ipaddr_string.split(/\s|,|;/).delete_if{ |y| y.empty? }
```
lnemsick-simp: Why are you allowing a string with space, ',', or ';' separated values?  That was not in the original function.  Also, if this functionality is required, you need to update the documentation to reflect this behavior and add a test for ',' separated and ';' separated values in a single string.

lib/puppet/functions/simplib/bracketize.rb
```diff
@@ -0,0 +1,63 @@
...
+  #
+  #   $bar = simplib::bracketize($foo)
+  #
+  #   $bar contains:[ '[2001:0db8:85a3:0000:0000:8a2e:0370:7334]',
+  #                   '[2001:0db8:85a3:0000:0000:8a2e:0370:7334]/24' ]
+  #
+  dispatch :bracketize do
+    required_param 'Array[String]', :ipaddr
+  end
+
```
lnemsick-simp: You need docs for each dispatch

lib/puppet/functions/simplib/bracketize.rb
```diff
@@ -0,0 +1,63 @@
...
+    bracketize(ipaddr)
+  end
+
+  def bracketize(ipaddr)
+    require 'ipaddr'
+    ipaddr = Array(ipaddr).flatten
+    retval = Array.new
+    ipaddr.each do |x|
+      begin
+        ip = IPAddr.new(x)
```
lnemsick-simp: Good idea. Much better way to figure out if IPv6 address.

lib/puppet/functions/simplib/bracketize.rb
```diff
@@ -0,0 +1,63 @@
+# Add brackets to IP addresses and `Arrays` of IP
+# addresses based on the rules for bracketing
+# IPv6 addresses.
+#
+# Ignore anything that does not look like an IPv6 address.
+#
+Puppet::Functions.create_function(:'simplib::bracketize') do
+
+  # @param ipv6 The ipv6 to bracketize
```
lnemsick-simp: This param much match the parameter in the dispatch.  In this case 'ipv6' -> ipaddr

spec/functions/simplib/bracketize_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+      ]
+      expected_output = [
+        '[2001:0db8:85a3:0000:0000:8a2e:0370:7334]',
+        '[2001:0db8:85a3:0000:0000:8a2e:0370:7334]/24'
+      ]
+
+      is_expected.to run.with_params(input).and_return(expected_output)
+    end
+
+    it 'returns same input if not an ipv6' do
```
lnemsick-simp: really is a test that returns input untouched if no transformation is required.  One of the entries is an IPv6 address that already has brackets

lib/puppet/functions/simplib/bracketize.rb
sharkbruhaha: roger, sorry was same way in original so wasn't aware.

lib/puppet/functions/simplib/bracketize.rb
sharkbruhaha: while `lib/puppet/functions/simplib/nets2ddq.rb` gives back an array, there are other cases were I could see people wanting to give a string straight to `bracketize` of just ipv6, so I figured I'd add the ability to do so since it was a simple addition.

Will update docs, thanks for the reminder

lib/puppet/functions/simplib/bracketize.rb
sharkbruhaha: roger

lib/puppet/functions/simplib/bracketize.rb
sharkbruhaha: thanks for the catch, was a holdover from when I was deciding on a name

PR #183, opened 2019-03-25
(SIMP-4891) Add simplib_networkmanager fact

trevor-vaughan:
This fact provides useful information about the state of NetworkManager
if it is present.

SIMP-4891 #comment Added simplib_networkmanager fact

### DIFF
--- a/lib/facter/simplib__networkmanager.rb
+++ b/lib/facter/simplib__networkmanager.rb
@@ -1,4 +1,4 @@
-Facter.add(:simplib_networkmanager) do
+Facter.add(:simplib__networkmanager) do
   confine :kernel => 'Linux'
 
   @nmcli_cmd = Facter::Util::Resolution.which('nmcli')
--- a/lib/facter/simplib_networkmanager.rb
+++ b/lib/facter/simplib_networkmanager.rb
@@ -1,0 +1,49 @@
+Facter.add(:simplib_networkmanager) do
+  confine :kernel => 'Linux'
+
+  @nmcli_cmd = Facter::Util::Resolution.which('nmcli')
+  confine { @nmcli_cmd }
+
+  setcode do
+    info = { 'enabled' => false }
+
+    nmcli_cmd = @nmcli_cmd + ' -t'
+
+    general_status = Facter::Core::Execution.execute(%(#{nmcli_cmd} -m multiline general status), :on_fail => :failed)
+    unless general_status == :failed
+      general_status = general_status.lines.map{|line| line.strip.split(':') }
+
+      info['enabled'] = true
+      info = {
+        'general' => {
+          'status' => Hash[general_status]
+        }
+      }
+    end
+
+    general_hostname = Facter::Core::Execution.execute(%{#{nmcli_cmd} general hostname}, :on_fail => :failed)
+    unless general_hostname == :failed
+      info['enabled'] = true
+      info['general'] ||= {}
+      info['general']['hostname'] = general_hostname.strip
+    end
+
+    connections = Facter::Core::Execution.execute(%(#{nmcli_cmd} connection show), :on_fail => :failed)
+    unless connections == :failed
+      info['enabled'] = true
+      info['connection'] = {}
+
+      connections.lines.each do |conn|
+        name, uuid, type, device = conn.strip.split(':')
+
+        info['connection'][name] = {
+          'uuid'   => uuid,
+          'type'   => type,
+          'device' => device
+        }
+      end
+    end
+
+    info
+  end
+end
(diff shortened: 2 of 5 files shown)


lib/facter/simplib_networkmanager.rb
```diff
@@ -0,0 +1,49 @@
...
+      info['general']['hostname'] = general_hostname.strip
+    end
+
+    connections = Facter::Core::Execution.execute(%(#{nmcli_cmd} connection show), :on_fail => :failed)
+    unless connections == :failed
+      info['enabled'] = true
+      info['connection'] = {}
+
+      connections.lines.each do |conn|
+        name, uuid, type, device = conn.strip.split(':')
```
lnemsick-simp: Will this parsing will do what you expect on the output from my dev machine? Notice the spaces such as 'Bridge br1':

$myhost: nmcli -t connection show
Bridge br1:2ee981ca-5ff4-4f9b-03fe-32879aa3dc85:bridge:br1
docker0:17f50a67-7276-4c44-b47c-4c3c58fea91b:bridge:docker0
System enp0s31f6:b5f59873-5bd2-c5a6-acb6-f03d5581aca5:802-3-ethernet:enp0s31f6
virbr0:dbdb93ed-c297-424a-9398-0c16700758c6:bridge:virbr0

op-ct: > Do you think the names with spaces in them will be useful?

@lnemsick-simp The "Bridge br1" string is the Connection Profile's [`id` string](https://developer.gnome.org/NetworkManager/stable/nm-settings.html), which is a "readable unique identifier for the connection" that is encouraged to contain spaces.  NetworkManager auto-generates the id (with the spaces) unless you change it yourself, so we can expect the spaces to be present most of the time.  

The `id` names are "useful" with `nmcli connection COMMAND <ID>` commands to serve as the `<ID>`.   From what I've seen, the `<ID>` in all `nmcli con` commands will also accept `uuid` or (DBUS) `path`.

op-ct: Reading @lnemsick-simp's question as "would something else make a more useful key?":  

The `id` is human-readable, but not optimal for manifest logic.  It's more likely to change than the (connection) `uuid`, and neither `id` or `uuid` describe anything any other Puppet facts refer to—such the device, which seems like the obvious key someone would want to correlate with this fact. 

On the face of it, it seems like we should either:
* use the device (`br1`) instead of the id the key, or 
* add an `info['device']` subtree with the output of `nmcli -t device`

Either solution would simplify the process of looking up connection information from a device name.

trevor-vaughan: @op-ct Unfortunately, if you want to actually use some of the `nmcli` commands, you need to use the name instead of the device from what I can tell.

trevor-vaughan: @op-ct I *think* that you can also have multiple entries pointing to the same device that are in different states of activation which is why you need the `name` as the key.

op-ct: @trevor-vaughan 

> @op-ct Unfortunately, if you want to actually use some of the `nmcli` commands, you need to use the name instead of the device from what I can tell.

That is precisely why either of the solutions above would result in a more useful Puppet fact:  

* No other Puppet facts or parameters report the NetworkManager connection `id` (or the interchangeable but similarly oblique `uuid` or (DBUS) `path`)—but they always report  a system's `interfaces`.   Puppet code trying to use this fact will almost certainly start with the interface (device) and need to map it to the connection in order to look up which one it needs.
  * If we used the device (`br1`) instead of the `id` as the key in the fact, or added an `info['device']` subtree to the fact structure (populated with the output of `nmcli -t device`), we could start with the device and just look it up.
  * Alternatively, adding an `info['device']` subtree to the fact structure (populated with the output of `nmcli -t device`) is still just two key lookups, which is more useful than coming up with your own data structure `dig()` logic in each place you reference the fact.
* Additionally, `id` does not make a very trustworthy identifier.  It is designed to be changed at any time (`nmcli con modify STRING1 connection.id STRING2`), unlike the device name.



> @op-ct I _think_ that you can also have multiple entries pointing to the same device that are in different states of activation which is why you need the `name` as the key.

According to the [`nmcli` man page](https://developer.gnome.org/NetworkManager/stable/nmcli.html), there can be multiple connections that apply to a device—but only one of them can be active on that device at any given time.

lib/facter/simplib__networkmanager.rb
```diff
@@ -0,0 +1,49 @@
...
+      info['general'] ||= {}
+      info['general']['hostname'] = general_hostname.strip
+    end
+
+    connections = Facter::Core::Execution.execute(%(#{nmcli_cmd} connection show), :on_fail => :failed)
+    unless connections == :failed
+      info['enabled'] = true
+      info['connection'] = {}
+
+      connections.lines.each do |conn|
```
jeannegreulich: The output for nmcli connection uses a space  deliminator not ':'

PR #185, opened 2019-04-05
(SIMP-4071) Re-enable simplib deprecation warnings by default

sharkbruhaha:
https://gitlab.com/simp/pupmod-simp-simplib/pipelines/55464984

### DIFF
--- a/lib/puppet/functions/simplib/deprecation.rb
+++ b/lib/puppet/functions/simplib/deprecation.rb
@@ -1,9 +1,6 @@
 # Function to print deprecation warnings, logging a warning once
 # for a given key.
 #
-# Messages can be enabled if the SIMPLIB_LOG_DEPRECATIONS
-# environment variable is set to 'true'
-#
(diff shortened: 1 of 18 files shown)


lib/puppet/functions/simplib/deprecation.rb
```diff
@@ -32,8 +29,6 @@ def deprecation(key, message)
       message = "#{message} at #{file}:#{line}"
     end
 
-    if ENV['SIMPLIB_LOG_DEPRECATIONS'] == 'true'
-      Puppet.deprecation_warning(message, key)
-    end
+    Puppet.deprecation_warning(message, key)
```
jeannegreulich: What do they do to turn off the deprecation warnings?

sharkbruhaha: they don't, they change their code to call the correct one, since this is the last release having these in here I felt being more heavy handed with it was justified and reasonable.

PR #191, opened 2019-05-21
(SIMP-6515) Defer to inbuilt fips_enable fact

trevor-vaughan:
Some version of facter have a built-in 'fips_enabled' fact. We now defer
to that one prior to adding our own fact.

Note: I was unable to determine how to properly unit test this but,
since the entire change is just checking for one item in Facter prior to
creating the fact, this isn't vastly different from the previous code
version (which also didn't have a unit test)

SIMP-6591 #close

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,6 @@
+* Tue May 21 2019 Trevor Vaughan <tvaughan@onyxpoint.com> - 3.15.1-0
+- Defer to the inbuilt 'fips_enabled' fact if it exists.
+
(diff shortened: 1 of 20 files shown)


lnemsick-simp:
https://gitlab.com/simp/pupmod-simp-simplib/pipelines/62472712

lnemsick-simp:
When I re-enabled the acceptance tests that had been inadvertently broken, they now fail.  Will fix.

lnemsick-simp:
:heavy_check_mark: https://gitlab.com/simp/pupmod-simp-simplib/pipelines/62479192

PR #192, opened 2019-05-31
(SIMP-MAINT) Prep for release - remove Puppet 4

lnemsick-simp:
- Removed Puppet 4 support, as Puppet has removed the Puppet 4 RPMs from
  their yum repos.  This means all the Puppet 4 acceptance tests fail.
- Pinned the version of stdlib to 5.2.0 in the .fixtures.yml, as the
  version on master is beyond the upper bound allowed in this
  component's metadata.json.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ addons:
 
 before_install:
   - rm -f Gemfile.lock
-  - for x in ${HOME}/.rvm/gems/*; do gem uninstall -I -x -i "${x}" -v '>= 1.17' bundler || true; gem uninstall -I -x -i "${x}@global" -v '>= 1.17' bundler || true; done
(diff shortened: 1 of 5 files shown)


lnemsick-simp:
https://gitlab.com/simp/pupmod-simp-simplib/pipelines/64163862

.travis.yml
```diff
@@ -38,7 +36,8 @@ addons:
 
 before_install:
   - rm -f Gemfile.lock
-  - gem install -v '~> 1.16' bundler
+  - for x in ${HOME}/.rvm/gems/*; do gem uninstall -I -x -i "${x}" -v '>= 1.17' bundler || true; gem uninstall -I -x -i "${x}@global" -v '>= 1.17' bundler || true; done
```
jeannegreulich: When I ran this on my workstation, it stopped and waited for input because bundler version 1.17.1 and 1.17.3 were installed.  Should there be a --force option on this or do we not think it will ever have 2 versions of the bundler installed because it is a clean build?

PR #195, opened 2019-08-02
(SIMP-6921) Support pupptlabs/stdlib 6.x.

pillarsdotnet:


lnemsick-simp:
In order to add Puppet 6, we will have to remove all Puppet 3 API functions.  Once SIMP 6.4.0 is released, we can do this.

Note: Removing all Puppet 3 API functions will be a major version bump.  Then, all SIMP modules will have to be updated to allow the later version of simplib.  A new ticket will be created for that follow-on work.

pillarsdotnet:
So this is blocking the `6.4.0` release?

Or will the `6.4.0` release not be compatible with Puppet 6.x?

lnemsick-simp:
SIMP 6.4.0 is the release in which we officially tell folks the Puppet 3 API functions will be removed in the next release.   Until SIMP 6.4.0 is released, I don't want to remove those functions from simplib, in case there is a critical bug that is found in the final testing that needs to be fixed.  It's easier to defer removing the Puppet 3 API changes until after SIMP 6.4.0, then to have to back them out or branch from the last tag.

lnemsick-simp:
SIMP 6.4.0 does not officially support Puppet 6.  There are some modules that do support Puppet 6, but not all of them do and the Puppet 6 RPMs are not packaged in the ISO.

lnemsick-simp:
:hourglass_flowing_sand: https://gitlab.com/simp/pupmod-simp-simplib/pipelines/77363310

PR #197, opened 2019-08-14
(SIMP-6919) Remove unnecessary EPEL repos in node sets

lnemsick-simp:
SIMP-6919 #comment pupmod-simp-simplib


lnemsick-simp:
Enough of the pipeline demonstrates the EPEL repos are unnecessary:
https://gitlab.com/simp/pupmod-simp-simplib/pipelines/76423878

PR #198, opened 2019-08-15
(SIMP-6916) Fix simplib::simp_version on Windows

trevor-vaughan:
SIMP-6916 #comment Fix simplib:;simp_version on Windows


lib/puppet/functions/simplib/simp_version.rb
```diff
@@ -13,16 +13,22 @@
   def simp_version(strip_whitespace = false)
     retval = "unknown\n"
 
-    if File.readable?('/etc/simp/simp.version')
-    # TODO Figure out under what circumstances the version string is prefaced
-    # with 'simp-'. This is not true for SIMP 6.x
-      version = File.read('/etc/simp/simp.version').gsub('simp-','')
+    is_windows = closure_scope['facts']['osfamily'].downcase == 'windows'
```
lnemsick-simp: where is the spec test for the new functionality?

trevor-vaughan: @lnemsick-simp The module doesn't "officially" support Windows and I wasn't keen on dropping it in there at this point. Do you think that I should?

The associated SIMP PR certainly exercises it.

lnemsick-simp: Seems like you could write a test by setting the fact to 'windows' without having it in the supported OS list.  That said,  you have a weird mix of partial Windows support. Kind of 'smells' to me.

trevor-vaughan: @lnemsick-simp It was specifically added because it was a bug that was triggered by the `simp-simp` Windows acceptance test PR.

I'm actually OK saying that `simplib` supports Windows (it should). I'll add it and see what happens.

trevor-vaughan: Well, now I feel a bit silly. Apparently, Windows was added to the metadata.json in the past :-|.

Adding the test now.

trevor-vaughan: Ok, not adding the test. We don't have any generic SIMP facts collected for Windows nodes and it also looks like the `on_supported_os` function doesn't handle the whitespace in the Windows versions properly so will need to be patched upstream.

trevor-vaughan: Added [SIMP-7004](https://simp-project.atlassian.net/browse/SIMP-7004) to cover getting the facts added.

trevor-vaughan: https://github.com/simp/rubygem-simp-rspec-puppet-facts/pull/17

trevor-vaughan:
Tests won't pass until https://github.com/simp/rubygem-simp-rspec-puppet-facts/pull/17 is merged and the Gem is released.

trevor-vaughan:
Tests are now passing!

PR #200, opened 2019-08-29
(SIMP-7008) Remove Puppet 3 Functions

jeannegreulich:
- removed Puppet 3 functions that were deprecated in 6.4.0

SIMP-7008 #close
SIMP-6246 #comment simplib complete


lnemsick-simp:
I screwed something up with my push.  Will fix next week.

PR #202, opened 2019-09-09
(SIMP-6916) Added blacklist to assert_metadata

trevor-vaughan:
The assert_metadata function needed a blacklist function for profile
support, specifically surrounding Windows

SIMP-6916 #comment add blacklist for windows profiles


REFERENCE.md
```diff
@@ -650,8 +650,9 @@ Fails a compile if the client system is not compatible with the module's
 `metadata.json`
 
 #### `simplib::assert_metadata(String[1] $module_name, Optional[Struct[{
```
op-ct: :thinking:  Thinking to how this function could be used after this feature merges:  different portions of a given module might not support all OSes equally, so the function's `$module_name` parameter might be better off re-named as something more generic, like `$section`, `$identifier`, `$subject`, or `$name_of_thing_that_will_throw_a_fit_if_it_doesnt_like_your_os`.

trevor-vaughan: @op-ct This is literally the module that you have to go look in and is the `$module_name` variable in Puppet. I figured that this was the best hint that I could give :-/.

I guess we could add a 'notes' section at some point but that's getting pretty busy.

op-ct: > This is literally the module that you have to go look in and is the $module_name variable in Puppet

@trevor-vaughan that is literally true, but come simp/pupmod-simp-simp#217, the message _`OS 'Windows' version 'Server 2012 R2' is not supported by module 'simp'`_ will be literally *un*true—the simp module does support Windows (in its scenario classes).  When a user follows up that error by looking in the module, its `metadata.json` will tell them the (literally) opposite story.

The misleading error message is really the problem, so we should at least alter the failure text to mention when the assertion fails due to blacklisting.  That was , users will be cued to expect support in other areas of the module.

functions/assert_metadata.pp
```diff
@@ -25,8 +27,9 @@
 function simplib::assert_metadata (
   String[1] $module_name,
   Optional[Struct[{
-    enable => Optional[Boolean],
-    os     => Optional[Struct[{
+    enable    => Optional[Boolean],
+    blacklist => Optional[Array[String]],
```
op-ct: It looks like `blacklist` can only prohibit entire OSes, but not specific versions.

functions/assert_metadata.pp
```diff
@@ -25,8 +27,9 @@
...
+    os        => Optional[Struct[{
```
op-ct: Should `blacklist` be part of the `os` structure instead of a separate top-level array?  That way it could take advantage of the existing logic, and fail on `none`, `full`, `major`.

functions/assert_metadata.pp
```diff
@@ -57,23 +61,26 @@ function simplib::assert_metadata (
...
+      fail("Could not find metadata for module '${module_name}'")
     }
 
     if $_options['os']['validate'] {
       if !$metadata['operatingsystem_support'] or empty($metadata['operatingsystem_support']) {
-        debug("'operatingsystem_support' was not found in '${module_name}'")
+        debug("'operatingsystem_support' was not found in module '${module_name}'")
+      }
+      elsif $facts['os']['name'] in $_options['blacklist'] {
+        fail("OS '${facts['os']['name']}' is not supported by module '${module_name}'")
```
op-ct: This "OS XXX is not supported by module YYY" message is misleading when the OS was blacklisted independently from the `metadata.json`; it should make the reason for the failure more specific in this case.

trevor-vaughan: @op-ct Can you think of a way to do this that's not going to require adding a ton of code everywhere?

functions/assert_metadata.pp
trevor-vaughan: I was thinking about this but it would basically have to be another identical hash.

trevor-vaughan:
@op-ct See if the latest changes work better for you

trevor-vaughan:
:hourglass_flowing_sand: https://gitlab.com/simp/pupmod-simp-simplib/pipelines/83908278

functions/module_metadata/assert.pp
```diff
@@ -0,0 +1,69 @@
+# Fails a compile if the client system is not compatible with the module's
+# `metadata.json`
+#
+# @param module_name
+#   The name of the module that should be checked
+#
+# @param options
+#   @see $simplib::module_metadata::blacklist::options
```
lnemsick-simp: Did you mean os_blacklist::options?

functions/module_metadata/assert.pp
```diff
@@ -0,0 +1,69 @@
...
+#
+# @return [None]
+#
+function simplib::module_metadata::assert (
+  String[1] $module_name,
+  Optional[Struct[{
+    enable => Optional[Boolean],
+    blacklist => Optional[Array[Variant[String[1], Hash[String[1], Variant[String[1], Array[String[1]]]]]]],
+    blacklist_validation => Optional[Struct[{
+      enable => Optional[Boolean],
```
lnemsick-simp: Why not just not specify the structure if you don't want to blacklist?  Seems confusing to specify the structure but turn off the capability.

functions/module_metadata/assert.pp
```diff
@@ -0,0 +1,69 @@
...
+# @param options
+#   @see $simplib::module_metadata::blacklist::options
+#
+# @return [None]
+#
+function simplib::module_metadata::assert (
+  String[1] $module_name,
+  Optional[Struct[{
+    enable => Optional[Boolean],
+    blacklist => Optional[Array[Variant[String[1], Hash[String[1], Variant[String[1], Array[String[1]]]]]]],
```
lnemsick-simp: Con't you want the blacklist Array size to be at least of length 1.  Otherwise the code below should check if the blacklist Array is empty.

functions/module_metadata/assert.pp
```diff
@@ -0,0 +1,69 @@
...
+#
+# @param options
+#   @see $simplib::module_metadata::blacklist::options
+#
+# @return [None]
+#
+function simplib::module_metadata::assert (
+  String[1] $module_name,
+  Optional[Struct[{
+    enable => Optional[Boolean],
```
lnemsick-simp: too many ways to enable/disable.  I would eliminate the global enable, if you really want individual enables

trevor-vaughan: @lnemsick-simp The global is for just killing the whole thing across the board. Otherwise you have to keep track of the (potentially updated) API as new capabilities get added.

functions/module_metadata/assert.pp
trevor-vaughan: Because Hiera stacking overrides. Taking something out of a deep_merged hash is pretty much impossible. Turning something off is easy.

functions/module_metadata/assert.pp
trevor-vaughan: Nah, an empty Array just returns false.

functions/module_metadata/assert.pp
lnemsick-simp: So you think the parameters called into the assert functions will be exposed by the class/define calling the assert function so that it can be overwritten in Hiera?    Sounds like you want the user to be able to override the check.  In other words you want the user to be able to say 'trust me, I know better  what you support' ?

lnemsick-simp: Why bother with the asserts if you allow the user to ignore?

functions/module_metadata/os_blacklisted.pp
```diff
@@ -0,0 +1,98 @@
...
+            'major': {
+              $_os_major_releases = $_os_versions.map |$os_release| { split($os_release, '\.')[0] }
+
+              $facts['os']['release']['major'] in $_os_major_releases
+            }
+            default: { true }
+          }
+        }
+        else {
+          true or $memo
```
lnemsick-simp: why even bother with $memo? true or anything will be true

functions/module_metadata/os_blacklisted.pp
```diff
@@ -0,0 +1,98 @@
+# Returns whether or not the passed module is blacklisted per the module's
+# metadata.json.
+#
+# If a blacklist is passed, then it will return `false` if the OS is in the
+# blacklist and `true` otherwise.
+#
+# If the blacklist is not matched, the standard OS validation will be used.
```
lnemsick-simp: Did I miss that logic?

functions/module_metadata/assert.pp
lnemsick-simp: Instead, why not expose a single variable to apply metadata.json checks or not and wrap the assert function call in if $check_metadata.  If you are dealing with a power user who wants to turn it off, keep it simple.  The power user assumes the full burden.

spec/functions/module_metadata/assert_spec.rb
```diff
@@ -0,0 +1,149 @@
...
+    end
+
+    context 'with an OS match' do
+      let(:facts){ valid_facts }
+
+      context 'with a simple list' do
+        it { expect { is_expected.to run.with_params('simplib', blacklist_base) }.to raise_error(/OS '#{facts[:os]['name']} #{facts[:os]['release']['full']}' is not supported at '/) }
+
+        context 'at the major version' do
+          it { expect { is_expected.to run.with_params('simplib', blacklist_major) }.to raise_error(/OS '#{facts[:os]['name']} #{facts[:os]['release']['full']}' is not supported at '/) }
```
lnemsick-simp: blacklist_major was merged with blacklist_complex.  So this is not a simple list test

spec/functions/module_metadata/assert_spec.rb
```diff
@@ -0,0 +1,149 @@
...
+        end
+
+        context 'at the full version' do
+          it { expect { is_expected.to run.with_params('simplib', blacklist_full) }.to raise_error(/OS '#{facts[:os]['name']} #{facts[:os]['release']['full']}' is not supported at '/) }
```
lnemsick-simp: blacklist_full was merged with blacklist_complex.  So this is not a simple list test

spec/functions/assert_metadata_spec.rb
```diff
@@ -1,114 +1,85 @@
 require 'spec_helper'
 
 describe 'simplib::assert_metadata' do
```
lnemsick-simp: Don't see any test for enable => false

spec/functions/module_metadata/assert_spec.rb
```diff
@@ -0,0 +1,149 @@
+require 'spec_helper'
```
lnemsick-simp: Don't see any tests for when global for specific checks are disabled.

spec/functions/module_metadata/os_supported_spec.rb
```diff
@@ -0,0 +1,143 @@
+require 'spec_helper'
```
lnemsick-simp: enable => false tests?

spec/functions/module_metadata/os_blacklisted_spec.rb
```diff
@@ -0,0 +1,79 @@
+require 'spec_helper'
```
lnemsick-simp: enable => false tests?

functions/module_metadata/assert.pp
trevor-vaughan: @lnemsick-simp This isn't for power users, this is for people that are throwing up their hands because things should work and don't want to read our code. One global switch to kill it all (just like we have now) and minor switches if you really know what you're doing (usually devs).

functions/module_metadata/os_blacklisted.pp
trevor-vaughan: Nah, that was a holdover from the 5th rewrite :-|

PR #203, opened 2019-09-24
(SIMP-7082) Added simplib__firewalls fact

trevor-vaughan:
Added a fact to gather the names of firewall software that can be
detected on the system.

Also:
  * Updated supported OS list
  * Added RHEL7 and RHEL8 as nodesets


spec/unit/facter/simplib__firewalls_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'spec_helper'
+
```
lnemsick-simp: What about iptables in the test?

trevor-vaughan: @lnemsick-simp It's just one from Linux, one from Solaris, and one from BSD.

The `pf` is specifically to make sure the command is translated properly.

Not too overly concerned in checking that hashes work if the data is correct.

jeannegreulich: What is the purpose of the fact then?  Can't you just set the firewall type data?

PR #204, opened 2019-09-30
(SIMP-5356) Update Passgen for BOLT

jeannegreulich:
- Bolt can't set the user to puppet/puppet.
- Updated the passgen function so it would not fail
  if the permissions on the passgen files could not get reset to
  puppet/puppet.

SIMP-5356 #close

### DIFF
--- a/lib/puppet/functions/simplib/passgen.rb
+++ b/lib/puppet/functions/simplib/passgen.rb
@@ -83,17 +83,20 @@ Puppet::Functions.create_function(:'simplib::passgen') do
     unless File.directory?(settings['keydir'])
       begin
         FileUtils.mkdir_p(settings['keydir'],{:mode => 0750})
-        # This chown is applicable as long as it is applied
-        # by puppet, not puppetserver.
-        FileUtils.chown(settings['puppet_user'],
-         settings['puppet_group'], settings['keydir']
-       )
       rescue SystemCallError => e
         err_msg = "simplib::passgen: Could not make directory" +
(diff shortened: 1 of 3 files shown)


lib/puppet/functions/simplib/passgen.rb
```diff
@@ -310,17 +318,23 @@ def lockdown_stored_password_perms(settings)
 
         unowned_files << file unless (file_owner == settings['puppet_user'])
```
lnemsick-simp: Line 319 will cause all the files to be removed when run in Bolt

lib/puppet/functions/simplib/passgen.rb
```diff
@@ -83,18 +83,21 @@ def passgen(identifier, modifier_hash=nil)
...
-         settings['puppet_group'], settings['keydir']
-       )
       rescue SystemCallError => e
         err_msg = "simplib::passgen: Could not make directory" +
          " #{settings['keydir']}:  #{e.message}. Ensure that" +
          " #{File.dirname(settings['keydir'])} is writable by" +
          " '#{settings['puppet_user']}'"
         fail(err_msg)
       end
+      begin
```
lnemsick-simp: Line 90  has a sentence 'Ensure that ...is writable by...' that is no longer applicable.

lnemsick-simp: Line 10 has a comment about failure related to puppet permissions that is not longer correct.

jeannegreulich:
Yeah, we aren't going to do this tis way.

PR #205, opened 2019-10-02
(SIMP-5356) Passgen in stand alone

jeannegreulich:
- Update the passgen function so it uses the process user id to check
    permissions on files instead of the puppet user.  This will allow
    stand alone processes to complete without error and still keep
    the checks for the puppet server working.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,7 @@
+* Wed Oct 02 2019 Jeanne Greulich <jeannegreulich@onyxpoint.com> - 4.0.0-0
+- Update passgen function to  set permissions on files to user
+  running the catalog compile.  This will allow bolt to set perms correctly.
+
 * Mon Sep 23 2019 Trevor Vaughan <tvaughan@onyxpoint.com> - 4.0.0-0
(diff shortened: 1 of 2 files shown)


trevor-vaughan:
This works as I expect on a live system. Tested by doing the following:

1. Install SIMP 6.4.0 via YUM
2. ``cp -r /usr/share/simp/environent-skeleton/puppet test_env``
3. ``cd test_env``
4. Add the following to ``data/default.yaml``:
```yaml
---
simp_options::puppet::server: "%{::fqdn}"
simp_options::puppet::ca: "%{::fqdn}"
simp_options::sssd: false
simp_options::pki: false
simp_options::dns::servers:
  - 8.8.8.8
clamav::rsync_source: ''

deferred_resources::mode: enforcing
deferred_resources::groups::install:
  administrators:
    members:
      - vagrant

deferred_resources::users::install:
  vagrant: ~

simp::classes:
  - simp_apache
```
5. Update the `simplib` module in `/usr/share/simp/modules`
6. Run: `/opt/puppetlabs/bin/puppet apply --hiera_config=env/hiera.yaml --modulepath=/usr/share/simp/modules/ env/manifests/site.pp`

PR #206, opened 2019-10-09
WIP (SIMP-7121) Use libkv in simplib::passgen

lnemsick-simp:
- Reworked `simplib::passgen` to use libkv for password persistence
- Added `libkv_options` parameter to `simplib::passgen`
- Updated `simplib::passgen` password options
  - Added `gen_timeout_seconds` option
  - Removed unnecessary test options: 'password' and 'salt'
- Updated spec and acceptance tests
- Updated spec_helper.rb

NOTE: This PR does not address migration of existing passwords from
previous versions of `simplib::passgen`.  That work will be addressed
in another ticket. This is why the PR is against the `libkv_integration`
branch instead of the `master` branch.

SIMP-7121 #close


lnemsick-simp:
https://gitlab.com/simp/pupmod-simp-simplib/pipelines/87747773

lnemsick-simp:
To support `simp passgen`, each password should have its `complexity` and `complex_only` options also stored, so that a regenerated password can be created matching the characteristics requested when simplib::passgen() was originally called.  (Length is easy to figure out...)

lnemsick-simp:
Going to rework this PR to support both the legacy and libkv-enabled `simplib::passgen`.  The user will opt into using libkv for `simplib::passgen` in hieradata.

PR #207, opened 2019-10-11
WIP (SIMP-7140) Add old password migration to libkv-enabled simplib::passgen

lnemsick-simp:
SIMP-7140 #close


lnemsick-simp:
This builds upon https://github.com/simp/pupmod-simp-simplib/pull/206

lnemsick-simp:
:heavy_check_mark: https://gitlab.com/simp/pupmod-simp-simplib/pipelines/88248207

lnemsick-simp:
Will cause flapping if the SIMP writable environment is shared (via links) among Puppet environments.  I can adjust for that (leaving the files intact), but will also have to add logic to handle password changes.  A lot of extra libkv calls for this case. Not ideal.

lnemsick-simp:
Does not make sense to review the current set of changes, as rewrite is necessary.

PR #208, opened 2019-10-15
(SIMP-7121) Added libkv mode to simplib::passgen

lnemsick-simp:
- Updated `simplib::passgen` to run in legacy mode or in a libkv mode.
  - When in libkv mode, `simplib:passgen` uses libkv for password persistence.
  - libkv mode is enabled by setting `simplib::passgen::libkv` to `true`
    in hieradata.
  - If you enable libkv mode in a system that already has passwords generated
    via the legacy code, currently, **all passwords will be regenerated**.
  - Added `libkv_options` parameter to `simplib::passgen` for use in libkv mode.
  - Added `gen_timeout_seconds` password option.  Previously this was hardcoded
    to 30 seconds.
- Added simp-libkv as a module dependency.

SIMP-7121 #close


lnemsick-simp:
https://gitlab.com/simp/pupmod-simp-simplib/pipelines/89039816

lnemsick-simp:
libkv doesn't really have any transaction support.  Perhaps we should store the current and previous password info at a single key instead of 2 keys.  Then, we won't have issues with partial info being stored, if  the previous password store succeeds but the current password store does not.

trevor-vaughan:
@lnemsick-simp Sounds sensible to me.

lnemsick-simp:
Want to rework the PR to avoid transaction problems.

lnemsick-simp:
https://gitlab.com/simp/pupmod-simp-simplib/pipelines/90305891

jeannegreulich:
Closing this because a later on that included these fixes was pushed.

PR #210, opened 2019-11-12
(SIMP-2827) more simplib changes for use in simp passgen

lnemsick-simp:
Updates required for `simp passgen`
- Added ability to set user and group for legacy passgen files.
  This is needed to ensure the owner can be set to puppet, when
  passwords are being modified/created via `puppet apply` operations
  executed as root. Otherwise, the files fail simplib::passgen
  validation.
- Added simplib::passgen::gen_salt function, so that a salt for
  the user-entered password can be generated in `simp passgen`
  via a `puppet apply`.

Other changes
- In simplib::gen_random_password, alternate use of character sets
  instead of lumping all characters from each character set together.
  When complexity > 0 and complex_only is false, this will generate
  a password that is more likely to pass SIMP's user password
  complexity requirements.

SIMP-2827 #comment pupmod-simp-simplib changes


lnemsick-simp:
https://gitlab.com/simp/pupmod-simp-simplib/pipelines/95398526

PR #211, opened 2019-11-20
(SIMP-7096) Add simplib::ip::family_hash function

trevor-vaughan:
Add `simplib::ip::family_hash` function that takes an IP address or
Array of IP addresses and returns a Hash with the addresses broken down
by family with additional helpful metadata

SIMP-7096 #comment Add simplib::ip::family_hash

### DIFF
--- a/lib/puppet/functions/simplib/ip/family_hash.rb
+++ b/lib/puppet/functions/simplib/ip/family_hash.rb
@@ -26,6 +26,35 @@ Puppet::Functions.create_function(:'simplib::ip::family_hash') do
   #         cidr: <CIDR netmask>
   #   ```
   #
+  # @example
+  #
+  #   simplib::ip::family_hash([
+  #     '1.2.3.4',
+  #     '2.3.4.5/8',
+  #     '::1'
+  #   ])
+  #
+  #   Returns (YAML Formatted for clarity)
+  #
+  #   ---
+  #   ipv4:
+  #     '1.2.3.4':
+  #       address: '1.2.3.4'
+  #       netmask:
+  #         ddq: '255.255.255.255'
+  #         cidr: 32
+  #     '2.3.4.5/8':
+  #       address: '2.0.0.0'
+  #       netmask:
+  #         ddq: '255.0.0.0'
+  #         cidr: 8
+  #   ipv6:
+  #     '::1':
+  #       address: '[::1]'
+  #       netmask:
+  #         ddq: nil
+  #         cidr: 128
+  #
   dispatch :family_hash do
     required_param 'Variant[
       Simplib::IP,
--- a/lib/puppet/functions/simplib/ip/family_hash.rb
+++ b/lib/puppet/functions/simplib/ip/family_hash.rb
@@ -1,0 +1,82 @@
+# Process an array of IP addresses and return them split by IP family and
+# include metadata and/or processed versions.
+#
+Puppet::Functions.create_function(:'simplib::ip::family_hash') do
+  # @param ip_addresses
+  #   The addresses to convert
+  #
+  # @return [Hash]
+  #   Converted Hash with the following format (YAML representation):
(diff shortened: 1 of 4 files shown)


lib/puppet/functions/simplib/ip/family_hash.rb
```diff
@@ -0,0 +1,82 @@
...
+Puppet::Functions.create_function(:'simplib::ip::family_hash') do
+  # @param ip_addresses
+  #   The addresses to convert
+  #
+  # @return [Hash]
+  #   Converted Hash with the following format (YAML representation):
+  #
+  #   ```
+  #   # IPv4 Addresses
+  #   ipv4:
```
lnemsick-simp: Why are these grouped by family?  Is there a case in which the user would prefer to have the family set in the individual entries?

lib/puppet/functions/simplib/ip/family_hash.rb
```diff
@@ -0,0 +1,82 @@
...
+  #       netmask:
+  #         ddq: <dotted quad notation netmask>
+  #         cidr: <CIDR netmask>
+  #   # IPv6 Addresses
+  #   ipv6:
+  #     <Passed Address>:
+  #       address: <normalized address>
+  #       netmask:
+  #         # DDQ is not valid for IPv6
+  #         ddq: nil
```
lnemsick-simp: why bother to set ddq if it is not applicable?

trevor-vaughan: Makes the dig logic easier when delving in using puppet language

lib/puppet/functions/simplib/ip/family_hash.rb
trevor-vaughan: Because this is a `family` hash. If someone wants something else, this isn't the data structure for them.

trevor-vaughan: Current usage is here: https://github.com/simp/pupmod-simp-iptables/pull/72/files#diff-57780fa042c02a1b6068fb93b67fd45dR163-R176

lib/puppet/functions/simplib/ip/family_hash.rb
```diff
@@ -0,0 +1,82 @@
...
+        Simplib::IP::V4::CIDR,
+        Simplib::IP::V6::CIDR
+      ]]
+    ]', :ip_addresses
+  end
+
+  def family_hash(addresses)
+    results = {}
+
+    Array(addresses).uniq.each do |addr|
```
lnemsick-simp: This eliminates duplicate input addresses (undocumented feature), but doesn't eliminate entries for which their metadata is the same.  For example, 1.2.0.0/16 is the same as 1.2.3.4/16.  Should the effective duplicates be removed as well?

lnemsick-simp:
Could create a puppet function that takes in an IP address/hostname and returns all transformations of it.  Then could use that in the different functions...

lib/puppet/functions/simplib/ip/family_hash.rb
trevor-vaughan: Honestly, I'm not sure. I wanted to leave it in for now to see if it would be an issue in the future. I almost used the normalized address as the key which would also solve that issue. @lnemsick-simp which way do you think it should go?

lnemsick-simp: I think this should return the normalized result.  All duplicates removed.  If the code you are writing doesn't care (the only user of the function at this point), I would definitely make the change.

trevor-vaughan: Turns out that I can't do this because the 'normalized' address is the bare address without the netmask (and needs to be that way to prevent additional re-splitting work later on). Reversing the two would make usage too difficult.

PR #212, opened 2019-12-04
(SIMP-7096) Fix simplib::ip::family_hash

trevor-vaughan:
simplib::ip::family_hash did not properly process 'unknown' entries

SIMP-7096 #comment fix simplib::ip::family_hash

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,7 @@
+* Wed Dec 04 2019 Trevor Vaughan <tvaughan@onyxpoint.com> - 4.0.1-0
+- Fixed an issue in `simplib::ip::family_hash` where the 'unknown' entries were
+  not properly populated.
+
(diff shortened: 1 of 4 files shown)


CHANGELOG
```diff
@@ -1,3 +1,7 @@
+* Wed Dec 04 2019 Trevor Vaughan <tvaughan@onyxpoint.com> - 4.0.1-0
+- Fixed an issue in `simplib::ip::family_hash` where the 'unknown' entries were
```
lnemsick-simp: Not really a bug fix.  You changed the function to handle hostnames (previously disallowed by parameter typing).

trevor-vaughan: @lnemsick-simp The reason that I think that it's a bugfix is that the `unknown` section of the hash was always supposed to be a catch-all. However, the way it was written, it couldn't actually function that way.

PR #214, opened 2020-02-26
(SIMP-7331) Renamspace of libkv to simpkv

trevor-vaughan:
* Requires https://github.com/simp/pupmod-simp-simpkv/pull/39

SIMP-7331 #close


lnemsick-simp:
⏳ Pushed to a local gitlab

PR #220, opened 2020-06-18
[SIMP-7822] Added additional helpers

trevor-vaughan:
- simplib::safe_filename() => Replace unsafe characters with safe ones in
  filename strings
- Simplib::Systemd::ServiceName => Data type for safe service name strings

SIMP-7822 #comment Additional helpers added

### DIFF
--- a/types/systemd/servicename.pp
+++ b/types/systemd/servicename.pp
@@ -1,0 +1,2 @@
+# Valid systemd service names
+type Simplib::Systemd::ServiceName = Pattern['^([A-za-z0-9.:_\\-]){1,256}$']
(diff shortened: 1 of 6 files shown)


types/systemd/servicename.pp
```diff
@@ -0,0 +1,2 @@
+# Valid systemd service names
+type Simplib::Systemd::ServiceName = Pattern['^([A-za-z0-9.:_\\-]){1,256}$']
```
ayohrling, line 2: Should this be capital Z here in the regex? i.e.

type Simplib::Systemd::ServiceName = Pattern['^([A-Za-z0-9.:_\\-]){1,256}$']

ayohrling, line 2:
```suggestion
type Simplib::Systemd::ServiceName = Pattern['^([A-Za-z0-9.:_\\-]){1,256}$']
```

PR #228, opened 2020-08-25
(SIMP-6740) simplib__auditd fact fix

trevor-vaughan:
- Fixed the simplib__auditd fact so that it detects the state of the running
  auditd process

SIMP-6740 #comment Fixed the simplib__auditd fact


op-ct:
(Resolved CHANGELOG conflict after merging #227)

PR #229, opened 2020-09-01
(SIMP-8228) Add simplib__mountpoints fact

trevor-vaughan:
- Add a simplib__mountpoints fact that returns information about mountpoints of
  specific interest to SIMP. This may be deprecated in the future if core facter
  can be updated to contain the relevant information.
- Refactor the tmp_mounts fact to use simplib__mountpoints
- Mark tmp_mounts as deprecated for removal in a future release
- Add a Simplib::Proc_hidepid() function that returns the numeric value of
  the `hidepid` option for the `/proc` mount and `0` by default

SIMP-8228 #comment add simplib__mountpoints fact


trevor-vaughan:
Back to the drawing board. I need the `gid` from `/proc` as the actual named group on the system

lib/facter/simplib__mountpoints.rb
```diff
@@ -0,0 +1,145 @@
...
+#
+# HELPERS
+#
+# * Any item starting with `_` has been added in as a 'helper' for ease in use
+#   inside of puppet code
+#
+# The following helpers are currently available:
+#
+#   * `_uid__user`  => User translation of the `uid` value
+#   * `_gid__group` => Group translation of the `gid` value
```
lnemsick-simp, line 20: This does not match the example below

lnemsick-simp (changes requested):
Essential changes:
* Fix documentation errata

Non-essential changes:
* Update to the latest Facter helper methods
* Provide unit tests for options that are in quotes to validate complex Regexes!

lib/facter/simplib__mountpoints.rb
```diff
@@ -0,0 +1,145 @@
...
+    known_translations = {}
+
+    # Check for bind mounts using findmnt since some systems (EL7) do not post the
+    # 'bind' keyword into the mount options any longer.
+    mount_list.each_key do |mnt|
+      mount_list[mnt]['options'] ||= []
+
+      unless mount_list[mnt]['options'].include?('bind')
+        findmnt_output = Facter::Util::Resolution.exec("findmnt #{mnt}")
+        # on RHEL 5 the command "findmnt" doesn't exist, if it doesn't exist
```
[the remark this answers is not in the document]
trevor-vaughan: I'll remove the comment, the code is pretty obvious

lib/facter/simplib__mountpoints.rb
```diff
@@ -0,0 +1,145 @@
...
+
+    target_dirs = ['/tmp', '/var/tmp', '/dev/shm', '/proc']
+
+    # Holder of the multi-call content
+    mount_list = {}
+
+    facter_mountpoints = Facter.value('mountpoints') || {}
+
+    # Sometimes Ruby has issues with /proc so fall back to a shell command
+    Facter::Util::Resolution.exec('cat /proc/mounts 2> /dev/null').each_line do |line|
```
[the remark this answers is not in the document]
trevor-vaughan: Yeah, I'll change that.

lib/facter/simplib__mountpoints.rb
```diff
@@ -0,0 +1,143 @@
...
+          mount_list[mnt]['options_hash']['_uid__user'] = found_uid
+
+          known_translations[:uid][mount_list[mnt]['options_hash']['uid']] = found_uid
+        rescue ArgumentError
+          # noop
+        end
+      end
+
+      next unless mount_list[mnt]['options_hash']['gid']
+
```
lnemsick-simp, line 128: Is there ever a case in which the `options_hash `would have the gid but not the uid?  If so, the `require 'etc'` would need to be in the gid-handling block, as well.

trevor-vaughan: Ah, I think this broke when I `rubocop'd` things

lnemsick-simp (approved):

PR #231, opened 2020-09-11
(SIMP-8382) Confine tmp_mounts on directories

trevor-vaughan:
SIMP-8382 #close

### DIFF
--- a/lib/facter/tmp_mounts.rb
+++ b/lib/facter/tmp_mounts.rb
@@ -21,8 +21,9 @@ simplib__tmp_mount_list = nil
 
 simplib__tmp_mount_target_dirs.each do |dir|
   Facter.add("tmp_mount#{dir.gsub('/','_')}") do
+    confine :kernel => :Linux
+    confine { File.directory?(dir) }
     setcode do
-      confine { File.directory?(dir) }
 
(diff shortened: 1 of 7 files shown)


lib/facter/tmp_mounts.rb
```diff
@@ -22,6 +22,8 @@
 simplib__tmp_mount_target_dirs.each do |dir|
   Facter.add("tmp_mount#{dir.gsub('/','_')}") do
     setcode do
+      confine { File.directory?(dir) }
```
lnemsick-simp, line 25: The original code had `confine :kernel => :linux`.  Is there a reason to switch to this directory confine?

lnemsick-simp: The original code only created the fact if the directory existed and then confined based on kernel.  Will just confining based on directory be sufficient?

trevor-vaughan: @lnemsick-simp Yep. Windows certainly won't have that directory.

trevor-vaughan:
Adding an acceptance test for Windows to make sure we don't blow it up unexpectedly again

lnemsick-simp (approved):

PR #232, opened 2020-09-15
Add facts to note EFI status

jcpunk:
It can be handy to see which hosts are booted into EFI and which are running secure-boot.


trevor-vaughan:
Single threading this into the conversation in #234

lib/facter/efi.rb
```diff
@@ -0,0 +1,11 @@
+# _Description_
+#
+# Return true if system booted via EFI
+#
+if Facter.value(:kernel).downcase == "linux" then
```
trevor-vaughan, line 5: Can this get changed to a `confine`?

lib/facter/secure_boot_enabled.rb
```diff
@@ -0,0 +1,25 @@
+# _Description_
+#
+# Return true if system booted via uEFI Secure Boot
+#
+if Facter.value(:kernel).downcase == "linux" then
```
trevor-vaughan, line 5: Can this get changed to a confine?

lib/facter/secure_boot_enabled.rb
```diff
@@ -0,0 +1,25 @@
...
+  Facter.add("secure_boot_enabled") do
+    setcode do
+      if File.exist?('/sys/firmware/efi')
```
trevor-vaughan, line 8: Moving this to a confine could be useful as well for readability:

`confine { File.exist?('/sys/firmware/efi') }`

lib/facter/secure_boot_enabled.rb
```diff
@@ -0,0 +1,25 @@
...
+        Dir.glob('/sys/firmware/efi/efivars/SecureBoot-*').each do | file |
+          File.open(file, 'r') do | hexcode |
+            hexcode.read(4)
+            code = hexcode.read(16).unpack('H*').first.to_i
+            if code == 1
```
trevor-vaughan, line 13: You can just leave this as `code == 1` and remove the `if/else` block since that will have the same result.

If you want to make this a bit cleaner, you could use a `retval`-style block where you set it to `false` by default and then set it to `code == 1` like in your #233 and #234 submissions.

trevor-vaughan (commented):
@jcpunk Thanks for this! It definitely looks useful. Have a few items that it would be great to get cleaned up if possible.

If you're up for adding tests, you can find examples in `spec/unit/facter`. If not, that's fine, we can pick it up from there.

If you're OK with me editing this PR a bit, let me know and I would be happy to make the updates and you can re-review before submission.

jcpunk:
I'm perfectly happy to have you make any edits you see fit.

As you can probably tell, ruby is not my forte....

lib/facter/secure_boot_enabled.rb
jcpunk: From a work flow perspective, wouldn't that prevent the fact from being defined on non-efi systems?  With facter considering `undef == true`, that feels like it would lead to false positives....

trevor-vaughan: In `facter` `undef` is `undef` and the Puppet language treats it as `false`.

We've trended towards using `dig` for pretty much everything like `if dig($facts, 'secure_boot_enabled') { ... }` and this takes care of any weird cases.

That said, I see your point. I think just moving to the `retval` style would help in terms of understanding that this fact should either be `true` or `false` and you're right that this should not be moved to a `confine` since that would be an unknown state.

trevor-vaughan:
OK I'll do a quick hack-around and you can see if it works for you

> 
> 
> I'm perfectly happy to have you make any edits you see fit.
> 
> As you can probably tell, ruby is not my forte....

trevor-vaughan:
@jcpunk See if this does what you expect.

jcpunk:
That looks good to me.  With the alternate names, it is probably worth getting README.md to match ;)

jcpunk:
I'd say this looks perfect :)

trevor-vaughan:
Great, going to throw in a quick test and we should be ready to roll

trevor-vaughan:
It looks like there are some edge cases with reading the files in `/sys/firmware/efi/efivars/SecureBoot-*`.

@jcpunk Do you have a reference site for the format of these files? Also, this is a loop, is stopping at the first valid match sufficient?

jcpunk:
I believe the first match of `/sys/firmware/efi/efivars/SecureBoot-*` should be sufficient.  At least, all my test systems only had one that matched.

My initial attempt to read the file was based in part on:
  https://github.com/lcp/mokutil/blob/master/src/mokutil.c#L1273
and some local testing.

but it looks like these[1] are probably the right place to look.  I think my provided reads look to be a bit too large...

[1]
https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/tools/testing/selftests/kexec/kexec_common_lib.sh?h=v5.4.65
https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/drivers/firmware/efi/libstub/secureboot.c?h=v5.4.65
https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/arch/x86/kernel/ima_arch.c?h=v5.4.65

trevor-vaughan:
@jcpunk Ah, spot on with those refs. It looks like we [need to actually check two files](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/tools/testing/selftests/kexec/kexec_common_lib.sh?h=v5.4.65#n52) (for whatever reason) and I do think your reads are a bit aggressive.

Would you mind giving it a refactor based on the self-test script? Matching what the devs do should always be right (probably :-D)

jcpunk:
I've reworked it with a note for where the upstream tests are located.

The kernel test seems to assume there is just one `/sys/firmware/efi/efivars/SecureBoot-*` and `/sys/firmware/efi/efivars/SetupMode-*` based on the `if -f` tests...

trevor-vaughan:
@jcpunk Thanks! I've got a case where the `SecureBoot` file exists but reading from it, even as root, is hit with an 'Access Denied' error. I'll wrap the fact to make sure that it doesn't explode on read errors. I'll leave it defaulting to `false` because an unknown status is `false`-ish.

trevor-vaughan:
@jcpunk Can you see if the latest push works for you?

jcpunk:
It returns the right results on my test systems.

trevor-vaughan:
Awesome! No way to stuff this through `beaker` that I know of so I just need to get some spec tests added and we'll be ready to roll!

trevor-vaughan (approved):

trevor-vaughan:
@jcpunk Just waiting for tests to finish before pushing. Thought you might want to take a look at the spec tests that were added for future reference.

jcpunk:
:) awesome thanks!

PR #233, opened 2020-09-15
Add a fact to track PCI devices

jcpunk:
At times it can be handy to search for M.2 disks or other hardware on the PCI bus.


trevor-vaughan:
Single threading this into the conversation in #234

PR #236, opened 2020-09-23
Add NUMA related facts

jcpunk:
This adds some structured facts to give insight into what NUMA support your hardware has.


lib/facter/simplib__numa.rb
```diff
@@ -0,0 +1,78 @@
...
+# _Description_
+#
+# Return supported NUMA possible nodes
+#
+Facter.add("simplib__numa_possible_nodes") do
+  confine :kernel => 'Linux'
+  confine { File.exist? '/sys/devices/system/node/possible' }
+
+  # read the file
+  File.open('/sys/devices/system/node/possible', 'r') do | text |
```
trevor-vaughan, line 24: This can be shortened to `File.read('/sys/devices/system/node/possible').strip` (you want to chop off leading and ending spaces

lib/facter/simplib__numa.rb
```diff
@@ -0,0 +1,78 @@
...
+    result = text.read()
+  end
+
+  setcode do
+    result
```
trevor-vaughan, line 29: Add the File read to the `setcode` block and make it the single File read and you don't need the `result` variable

lib/facter/simplib__numa.rb
```diff
@@ -0,0 +1,78 @@
...
+Facter.add("simplib__numa_online_nodes") do
+  confine :kernel => 'Linux'
+  confine { File.exist? '/sys/devices/system/node/online' }
+
+  # read the file
+  File.open('/sys/devices/system/node/online', 'r') do | text |
+    result = text.read()
+  end
+
+  setcode do
```
trevor-vaughan, line 47: Same `File.read()` change here and toss everything that's not a `confine` into the `setcode` for isolation.

lib/facter/simplib__numa.rb
```diff
@@ -0,0 +1,78 @@
...
+
+# _Description_
+#
+# Return NUMA memory per node
+#
+Facter.add("simplib__numa_node_memory") do
+  confine :kernel => 'Linux'
+  confine { '/sys/devices/system/node/possible' }
+
+  result = {}
```
trevor-vaughan, line 61: Pop the non-confine bits into the `setcode` for block isolation please.

lib/facter/simplib__numa.rb
```diff
@@ -0,0 +1,78 @@
...
+
+  Dir.glob('/sys/devices/system/node/node*').each do | file |
+    begin
+      File.foreach(file + '/meminfo') do | text |
+        if text =~ / MemTotal:/
+          kb = text.split(':')[1].tr("^0-9", '')
```
trevor-vaughan, line 67:
```suggestion
          kb = text.split(':')[1].delete("^0-9")
```

lib/facter/simplib__numa.rb
```diff
@@ -0,0 +1,78 @@
...
+          bytes = kb * 1024
```
trevor-vaughan, line 68: I don't think this is what you want.

The result of `kb` will be a String. A String * Integer == that String repeated Integer times.

You probably want `kb.to_i * 1024`.

trevor-vaughan (changes requested):
@jcpunk Nice! Checking to see how much load this would give but it doesn't seem unreasonable at first glance.

Would you be OK consolidating all of this into a single `simplib__numa` Hash though? It's not really required but trends towards the modern `facter` way of collecting up related facts (`os`, `network`, etc...)

jcpunk:
I'll slam these all into a single structured fact and in theory also address the changes requested shortly.

lib/facter/simplib__numa.rb
```diff
@@ -0,0 +1,33 @@
...
+# and a memory mapping layout
+#
+Facter.add("simplib__numa") do
+  confine :kernel => 'Linux'
+  confine { File.exist? '/sys/devices/system/node' }
+
+  setcode do
+    result = {}
+
+    # read the 'possible' nodes file
```
trevor-vaughan, line 13: Nice!

Sorry that I didn't think about this earlier, but you may want to add `if File.exist?('<whatever>')` to these just so you don't throw a weird error.

I mean they *should* be there but :-|.

jcpunk: Just added now :)

trevor-vaughan:
@jcpunk Looks good! Are you up for trying your hand at writing tests? If not, a merge will have to wait until someone has time to pop them in.

The tests for the secure boot update can be found at https://github.com/simp/pupmod-simp-simplib/blob/master/spec/unit/facter/simplib__secure_boot_enabled.rb

jcpunk:
I can take a stab at the tests.  How would I fake out the `confine { File.exist? '/sys/devices/system/node' }` ?

trevor-vaughan:
@jcpunk Something like this: https://github.com/simp/pupmod-simp-simplib/blob/master/spec/unit/facter/simplib__sshd_config_spec.rb#L11

jcpunk:
In theory there are now tests....

trevor-vaughan (changes requested):
Some updates need to be done to the tests.

Blocking and happy to fix things up since valid examples have been provided.

jcpunk:
I'll confess I'm not sure what to do from here...

trevor-vaughan:
@jcpunk No problem! I'll work on getting it in shape over the weekend. I noticed that there are going to be a few `rspec` nuances in there so happy to help. Thanks for giving it a stab though. Having something that's mostly there helps a lot (particularly the file dumps)

trevor-vaughan:
@jcpunk Please see if the latest version works for you. I had to make some assumptions because I couldn't go back and look at the code history in the ticket.

If possible, try not to force push so that old references are held.

trevor-vaughan:
@jcpunk I also just realized that it might be handy to have a `nodes` key in the hash that is just an Array of all discovered nodes. This makes looping easier than having to extract keys and do pattern matches.

jcpunk:
In theory I've added a list of known nodes to the facts produced.

lib/facter/simplib__numa.rb
```diff
@@ -0,0 +1,41 @@
...
+    if File.exist?('/sys/devices/system/node/online')
+      result['online'] = File.read('/sys/devices/system/node/online').strip
+    end
+
+    require 'pathname'
+    Dir.glob('/sys/devices/system/node/node*').each do | file |
+      meminfo_file = Pathname.new(File.join(file, 'meminfo'))
+      next unless meminfo_file.exist?
+      nodename = File.basename(file)
+      result['nodes'].append(nodename)
```
trevor-vaughan, line 29: Oh Python :-D
```suggestion
      result['nodes'].push(nodename)
```

spec/unit/facter/simplib__numa_spec.rb
```diff
@@ -0,0 +1,321 @@
...
+            Node 1 HugePages_Surp:      0
+            EOMEM
+        end
+      end
+
+      it do
+        expect(Facter.fact('simplib__numa').value).to eq({
+          'possible' => '0-1',
+          'online'   => '0',
+          'nodes'    => ['node0'],
```
trevor-vaughan, line 240:
```suggestion
          'nodes'    => ['node0', 'node1'],
```

trevor-vaughan (changes requested):
@jcpunk Just a few items and it should be good to go!

trevor-vaughan (approved):

PR #237, opened 2020-10-23
(SIMP-8590)  Fix simplib::debug::inspect with Bolt

trevor-vaughan:
Fixed:
* Bolt does not set scope.source. The code was reworked to optionally
  pull scope.source and simply insert the scope as a string as a
  fallback.

SIMP-8590 #close


op-ct (approved):
👁️ 👍

PR #239, opened 2020-11-15
(SIMP-7822) Add simplib__package_managers

trevor-vaughan:
- Added simplib__package_managers fact
- Fixed names of varous facter spec tests

SIMP-7822 #comment add simplib__package_managers fact


jeannegreulich:
There is a fact called package_provider in stdlib that we are going to use instead.
I am going to take this PR and pull out the SIMP-MAINT stuff.

PR #252, opened 2021-04-29
(SIMP-9427) Add simplib::params2hash()

trevor-vaughan:
Added:
  - Add a `simplib::params2hash()` function to return all of the calling scope's
    parameters as a Hash

SIMP-9427 #comment added simplib::params2hash() to make sssd management easier

### DIFF
--- a/lib/puppet/functions/simplib/params2hash.rb
+++ b/lib/puppet/functions/simplib/params2hash.rb
@@ -1,4 +1,7 @@
 # Returns a Hash of the parameters of the calling resource
+#
+# This is meant to get the parameters of classes and defined types.
+# The behavior when calling from other contexts is undefined
 Puppet::Functions.create_function(:'simplib::params2hash', Puppet::Functions::InternalFunction) do
 
   # @param prune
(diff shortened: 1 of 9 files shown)


spec/spec_helper_acceptance.rb
```diff
@@ -1,16 +1,11 @@
+# frozen_string_literal: true
+
 require 'beaker-rspec'
 require 'tmpdir'
 require 'yaml'
 require 'simp/beaker_helpers'
 include Simp::BeakerHelpers
```
lnemsick-simp, line 8: There is still a windows acceptance test suite and it is still being executed in GitLab.  Isn't the Windows configuration required for that test.

lnemsick-simp: That test successfully ran a month ago.

trevor-vaughan: @op-ct  I did a diff between this and the last passing Windows tests and I don't see anything that would have caused an issue. Could you take a look?

lib/puppet/functions/simplib/params2hash.rb
```diff
@@ -0,0 +1,26 @@
+# Returns a Hash of the parameters of the calling resource
```
lnemsick-simp, line 1: Might be worthwhile noting this is intended to be used with classes and defines. When used with a puppet language function returns {} because of the scope search.

lnemsick-simp (approved):
Problem with windows test in GitLab is known and managed by a separate ticket.

PR #264, opened 2021-09-23
(SIMP-10456) Better password randomization

trevor-vaughan:
- Fixed
  - Increased randomization in simplib::gen_random_password

SIMP-10456 #comment better password randomization for FIPS mode

### DIFF
--- a/lib/puppet/functions/simplib/gen_random_password.rb
+++ b/lib/puppet/functions/simplib/gen_random_password.rb
@@ -74,6 +74,8 @@ Puppet::Functions.create_function(:'simplib::gen_random_password') do
           rand_list_index = rand_list_index-1
         end
 
+        last_list_rand = rand_list_index
+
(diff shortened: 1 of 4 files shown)


lib/puppet/functions/simplib/gen_random_password.rb
```diff
@@ -31,44 +31,64 @@ def gen_random_password(length, complexity=nil, complex_only=false, timeout_seco
...
       end
 
-      index = 0
+      last_list_rand = nil
+      last_char_rand = nil
       Integer(length).times do |i|
-        passwd += charlists[index][rand(charlists[index].length-1)]
-        index += 1
-        index = 0 if index == charlists.length
+        rand_list_index = rand(charlists.length).floor
```
lnemsick-simp, line 71: when the arg passed to rand is an Integer, it returns Integers.  So, the floor is not actually needed

lnemsick-simp (approved):

PR #266, opened 2021-12-16
klist -> klist -s

ke5C2Fin:
Change klist to klist -s to so that it returns 1 when credentials cache cannot be read or is expired, thereby triggering the following kinit.

el7, el8, simp/sssd, simp/simp_ipa
Without this the ipa fact fails and the simp/sssd module removes the ipa domain and provider files if the credential cache is expired.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,7 @@
+* Thu Dec 16 2021 ke5C2Fin <noreply@github.com> - 4.10.2
+- Fixed
+  - Call `klist -s` instead of `klist` to properly handle cache issues
+
 * Thu Sep 23 2021 Trevor Vaughan <tvaughan@onyxpoint.com> - 4.10.1
 - Fixed
   - Increased randomization in simplib::gen_random_password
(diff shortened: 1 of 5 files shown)


trevor-vaughan:
@ke5C2Fin Since this is on your `master` branch, are you OK with me updating the `CHANGELOG` or would you like to do it yourself?

ke5C2Fin:
Go for it.
I will make sure any future PR's are appropriately from a branch.

trevor-vaughan (approved):

op-ct (approved):
For anyone else curious why running with `-s` (silent) is needed for klist to properly handle cache issues:

> `-s`
>
> causes klist to run silently (produce no output), but to still set the exit status according to whether it finds the credentials cache. **The exit status is '0' if klist finds a credentials cache, and '1' if it does not or if the tickets are expired.**

op-ct:
Re-ran failed acceptance tests following the release of simp-beaker-helpers [1.24.1][1]; all tests now pass

[1]: https://github.com/simp/rubygem-simp-beaker-helpers/releases/tag/1.24.1

PR #288, opened 2023-08-28
Ruby 3 support

silug:
* Support stdlib 9
* Update gem dependencies


rgardner4012 (approved):

kjetilho:
this broke Ruby 2.4 and earlier, since you can't do `rescue` in a `do` block.  to make it work on any Ruby version, the internals of the `do` block for `setcode` must be wrapped like  `setcode do begin ... rescue ... end end`.  here's the reference to when the Ruby grammar was changed:

https://bugs.ruby-lang.org/issues/12906

